PR #4, opened 2015-09-03
Feature/woo commerce2.4 api key

0x5378:
Included are the requested changes.

They have been tested to the best of my abilities but I would like to go over them again tomorrow with fresh eyes.

I mistyped one of the issue numbers in a commit which may cause some confusion.

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -67,8 +67,8 @@ class WC_Taxjar_Integration extends WC_Integration {
       add_filter( 'woocommerce_customer_taxable_address', array( $this, 'append_base_address_to_customer_taxable_address' ), 10, 1 );
  
       // WP Hooks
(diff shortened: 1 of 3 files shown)


tonkapark:
Hi -

Great first pass, a few additional tweaks/suggestions.

Let's change the default message when the taxjar_download checkbox is unchecked on version 2.4 to read differently, "If enabled, you can then generate secure tokens to grant TaxJar access to your store to download your order history."

When there are no 2.4 keys created the wording should be,
"You have not yet granted access toTaxJar for order downloads, Generate WooCommerce API Key now. <a>Further instructions</a>"

Instructions link is http://support.taxjar.com/knowledge_base/topics/how-to-generate-woocommerce-2-dot-4-api-keys

If there already exists a 2.4 token with a user name like *taxjar* or the description of they key is like TaxJar then we should show that the key exists and a link to the API tab to manage them.

Thanks

PR #10, opened 2015-12-18
Update connection messaging

0x5378:
Broke connection, account, and api verification out into a new class.

Only show TaxJar Status area if there is an issue with connection, api key, or account.


tonkapark:
:+1:

PR #11, opened 2015-12-29
Hotfix for shipping tax calculations

0x5378:
Fixed a simple math bug that was causing an error in displaying the correct tax amount on checkout page.

Added a new spec to prove the bug and fix.


fastdivision:
@0x5378 With more specs in the pipeline can we get Travis CI up and going for this repo? https://github.com/woothemes/woocommerce/blob/master/.travis.yml

fastdivision:
Fix looks good to me btw :+1:

PR #14, opened 2016-06-27
properly handle when shipping is disabled in WooCommerce

0x5378:
Hi all,

This PR contains a fix for an API error that was being raised when shipping is disabled.  I also changed the connection code to use home_url() rather than site_url().  This has been tested by completing orders in the following scenarios:

- Purchased a physical product with shipping turned off.
- Purchased a virtual product which has shipping disabled.
- Purchased a physical product with shipping turned on.

I tested this using the latest version of WordPress (4.5.3) and the latest version of WooCommerce (2.6.1)


tonkapark:
👍

Issue #21, opened 2017-05-18
to_street is not being sent for SSG state calculations


bradmkjr:
Currently, SSG states require a street address to return a correct tax rate, otherwise a lower rate is returned. The correct rate is calculated for tax payments, resulting in a under collection at time of purchase. Needs to add to_street parameter to the api request.

fastdivision:
Finished in #75

PR #22, opened 2017-05-18
Adds Street Address to criteria for calculating tax rates

bradmkjr:
Closes issue #21 by sending the address to the API to calculate the correct tax rate for SSG.

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -54,6 +54,8 @@ class WC_Taxjar_Integration extends WC_Integration {
       // Calculate Taxes
       add_action( 'woocommerce_calculate_totals', array( $this, 'use_taxjar_total' ), 20 );
 
+      // Removed in WooCommerce 3.0
+      // https://wordpress.org/support/topic/woocommerce-3-0-missing-hook-woocommerce_ajax_calc_line_taxes/
       add_filter( 'woocommerce_ajax_calc_line_taxes', array( $this, 'admin_ajax_calculate_taxes' ), 99, 4 );
 
       // Settings Page
@@ -274,6 +276,7 @@ class WC_Taxjar_Integration extends WC_Integration {
         'to_state' =>         null,
         'to_zip' =>           null,
         'to_city' =>          null,
+        'to_street' =>        null,
         'amount' =>           null, // $this->taxjar_taxable_amount($woocommerce->cart);
         'shipping_amount' =>  null // $woocommerce->shipping->shipping_total
     ), $options) );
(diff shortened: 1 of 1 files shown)


includes/class-wc-taxjar-integration.php
```diff
@@ -575,7 +636,7 @@ public function admin_ajax_calculate_taxes( $items, $order_id, $country, $post )
   */
   public function append_base_address_to_customer_taxable_address( $address ){
     $store_settings = $this->get_store_settings();
-    list( $country, $state, $postcode, $city ) = $address;
+    list( $country, $state, $postcode, $city, $street ) = $address;
```
fastdivision: Let's use `array_pad( $address, 5, '' )` to avoid undefined offset errors.

includes/class-wc-taxjar-integration.php
```diff
@@ -435,6 +450,7 @@ public function taxjar_api_call( $options = array() ) {
       $tax_rates = array(
 			  "tax_rate_country" =>   $to_country,
 			  "tax_rate_state" =>     $to_state,
+        "tax_rate_street" =>    $to_street,
```
fastdivision: This is causing a database error since the column doesn't exist when creating / updating a tax rate.

includes/class-wc-taxjar-integration.php
```diff
@@ -450,6 +466,7 @@ public function taxjar_api_call( $options = array() ) {
 			$wc_rates = WC_Tax::find_rates( array(
         'country' =>        $to_country,
         'state' =>          $to_state,
+        'street' =>         $to_street,
```
fastdivision: Unfortunately we can't look up rates by street address yet using `find_rates`.

fastdivision:
Hi @bradmkjr, thanks for working on this update! From my initial review, it looks like we'll have to do some additional work if we want to store and reference tax rates at the street-level. I'm seeing the `to_street` parameter hit our API after bypassing a couple errors, but the rate stored in the database is still tied to a city and postal code at most. This rate won't reflect street-level, rooftop accurate rates returned by our API if the street happens to change.

I'm thinking we'd have to override the `_update_tax_rate_locations` method to store the street in the `tax_rate_locations` table, but we'd also have to override Woo's methods for looking up a tax rate like `find_rates`. I'll see what @0x5378 thinks next week.

In the meantime, I commented on the parts of your code causing errors with debug mode enabled. Appreciate the help!

fastdivision:
Closing this PR in favor of https://github.com/taxjar/taxjar-woocommerce-plugin/pull/33 since it's more up-to-date with recent changes. Appreciate the help!

PR #32, opened 2017-07-21
Fix Subscription Calculations at Checkout

fastdivision:
This PR will only calculate tax on subscriptions when a recurring payment occurs. If a subscription has a trial period, the line subtotal is $0. Check both the unit price **and** line subtotal before adding line items to an API calculation.


matthewecampbell:
Looks good to me.  I tested this out locally and it ran well.  It works when subscriptions are entered alone or paired in a cart with a non-subscription item.  It looks ready to go!

PR #36, opened 2017-09-06
WooCommerce 3.2 Support

fastdivision:
WooCommerce 3.2 brings coupons / discounts to backend orders and a "Store Address" section to Settings > General.

For now, we're supporting backend order discounts and considering the additional "Store Address" fields later. State and country should still use the `woocommerce_default_country` option.

**Versions Tested:**

- [x] Woo 3.2 Beta
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6.2
- [x] Woo 2.6.1
- [x] Woo 2.6.0


fastdivision:
Specs passing in all versions, front-end and back-end calculations working fine. There are a couple issues around WooCommerce 3.2 and WooCommerce Subscriptions, but not directly tied to TaxJar:

- Subscriptions plugin causes issues with backend order coupons: "Sorry, this coupon is only valid for an initial payment and the cart does not require an initial payment."

- Subscription line items with a trial period return unit price and line subtotals.

We'll see if WooCommerce Subscriptions is updated accordingly for 3.2.

matthewecampbell:
Tested everything out locally.  It seems to work beautifully on our end.

Calculations work on the backend and front end.  One small thing that I see you noted is that Subscriptions works a bit weird with a trial period.  The trial seems to be ignored, but it doesn't seem to be our issue.

Green light from me!

PR #39, opened 2017-10-21
Calculate order totals twice for WooCommerce 3.2

fastdivision:
WooCommerce 3.2 completely revamped their tax calculation process and changed the order on how they calculate item totals, shippings, and fees. The `woocommerce_calculate_totals` action now comes at the end of this process.

In order to properly calculate the order total, we need tax rates in the system to be applied to the item totals. However, we can't get the amount to collect without calculating discounts first. So if we try to hook only into `woocommerce_before_calculate_totals`, we won't be able to properly calculate the total amount of sales tax to collect. If we only hook into `woocommerce_calculate_totals`, the new rate won't be applied to the item totals and we won't update the grand total correctly.

For this reason, we need to calculate the order totals twice. The first time allows us to create a new tax rate in WooCommerce using `woocommerce_before_calculate_totals`. The second time we actually apply the tax rate and get the correct amount to collect with discounts applied using `woocommerce_calculate_totals`.

👍

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -44,6 +44,7 @@ class WC_Taxjar_Integration extends WC_Integration {
 		if ( ( 'yes' == $this->settings['enabled'] ) ) {
 
 			// Calculate Taxes at Cart / Checkout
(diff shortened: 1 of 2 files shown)


fastdivision:
Closing this PR in favor of a simpler approach, see pull request #40.

PR #41, opened 2017-11-13
Always Use Taxable Address for WC Rate Lookups

fastdivision:
If a merchant has origin-based or CA modified-origin sourcing, our plugin could find a different native tax rate than WooCommerce's destination-based lookup since we were looking at the store's city and zip code, not the customer's city and zip code. We should always use the taxable address for looking up rates, similar to WooCommerce, to ensure we always update the same rate WooCommerce is using for internal total calculations. Using the same rate ensures the grand total always reflects the subtotal, shipping, and tax returned by TaxJar.

**Versions Tested:**

- [x] Woo 3.2
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6.2
- [x] Woo 2.6.1
- [x] Woo 2.6.0

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -280,7 +280,6 @@ class WC_Taxjar_Integration extends WC_Integration {
 		$this->freight_taxable      = 1;
 		$this->line_items           = array();
 		$this->has_nexus            = 0;
-		$this->tax_source           = 'origin';
 		$this->rate_ids             = array();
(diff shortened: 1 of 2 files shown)


matthewecampbell:
The update looks good to me.

Tested on 

- [ ] 2.6.0

- [ ] 2.6.1

- [ ] 2.6.2

- [ ] 3.0

- [ ]  3.1

- [ ]  3.2.3

PR #42, opened 2017-11-14
Recalculate totals in WooCommerce 3.2 instead of manually updating total

fastdivision:
We've seen various edge cases with 3rd party plugins manipulating totals, such as order-level discounts causing issues with how TaxJar adds sales tax to the grand total.

Instead of using `woocommerce_calculated_total`, leverage `woocommerce_after_calculate_totals` in WooCommerce 3.2 and re-calculate for new tax rates introduced in the system.

**Versions Tested:**

- [x] Woo 3.2
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6


matthewecampbell:
Tested in Woo 2.6
Woo 2.6.1
Woo 2.6.2
Woo 3.0
Woo 3.2.3

Everything looks good!  Calculations are working properly.

PR #46, opened 2017-11-29
Exempt FL Shipping Tax If Local Pickup Available

fastdivision:
If a local pickup shipping method is shown to the customer in Florida, we should exempt shipping tax: https://blog.taxjar.com/shipping-in-florida-taxable/

This PR sets the shipping tax to $0 and turns off freight taxability so the tax rate saved doesn't affect shipping, which hopefully prevents the shipping tax from being added after processing the payment and saving the order.

**Versions Tested:**

- [x] Woo 3.2
- [ ] Woo 3.1
- [ ] Woo 3.0
- [ ] Woo 2.6.14
- [ ] Woo 2.6.2
- [ ] Woo 2.6.0

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -355,6 +355,12 @@ class WC_Taxjar_Integration extends WC_Integration {
 			}
 
 			$this->item_collectable = $this->amount_to_collect - $this->shipping_collectable;
+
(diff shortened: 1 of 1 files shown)


fastdivision:
Closing this PR until we introduce end-to-end shipping taxability overrides for TaxJar Reports and AutoFile when importing WooCommerce transactions into TaxJar.

PR #48, opened 2017-12-02
Fix Potential Logging Issue on Shared Hosting

fastdivision:
If a TaxJar user has "Debug Log" enabled in their configuration on a shared hosting plan, it's possible to receive "too many open files" server errors. Legacy code in our plugin was instantiating the WC_Logger for every `_log()` call which is bad. This PR ensures we only instantiate WC_Logger once if it isn't already set.

**Versions Tested:**

- [x] Woo 3.2
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6.14
- [x] Woo 2.6.2
- [x] Woo 2.6.0

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -220,7 +220,9 @@ class WC_Taxjar_Integration extends WC_Integration {
 			}
 		}
 		if ( $this->debug ) {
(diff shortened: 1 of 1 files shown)


fastdivision:
No issues here on 2.x or 3.x. Minor fix, going to merge 👍

PR #52, opened 2017-12-08
Rounding Fix for Line Item Rounding in WC 3.2

fastdivision:
By default the TaxJar plugin disables "Round tax at subtotal level":

screen shot 2017-12-08 at 3 12 15 pm

When we set the tax total in the cart, we should respect this setting and sum up the tax by line item instead. However, this change alone doesn't resolve rounding issues on the grand total. WC handles that calculation for us.

WC 3.2.5 wasn't properly rounding per line, so the spec attached to this PR won't pass. After [merging this PR](https://github.com/woocommerce/woocommerce/pull/18008), the spec passes and the rounding issue appears to be fully resolved. So once WC 3.2.6 or 3.3 is shipped we'll get this merged into the TaxJar plugin.

**Versions Tested:**

- [x] Woo 3.2
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -280,7 +280,6 @@ class WC_Taxjar_Integration extends WC_Integration {
 
 		$this->tax_rate             = 0;
 		$this->amount_to_collect    = 0;
-		$this->item_collectable     = 0;
 		$this->shipping_collectable = 0;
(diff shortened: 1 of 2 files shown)


matthewecampbell:
Works well on 3.2.6! Need to do further testing on 3.2 and below.

PR #59, opened 2018-02-15
Moved the total price calulation until after the taxes are added

### LINKED ISSUE #58: Total price at the checkout not calculated correctly if shipping method is not taxable
Related issue in WooCommerce Services: https://github.com/Automattic/woocommerce-services/issues/1309

To reproduce:
* use WooCommerce 3.3.1
* set up a store in a state that taxes shipping (for example NC)
* add a Flat Rate shipping method and set its taxable status to None
* checkout a product
* the tax row shows the tax for both the item and shipping but the total only includes the item tax:
screen shot 2018-02-15 at 19 49 34

In this scenario either the taxable setting on the shipping method should be ignored or the tax row should not include shipping tax.

CC @jeffstieler

marcinbot:
Fixes #58 

By moving `new WC_Cart_Totals( $wc_cart_object );` until after the taxes are set on `$wc_cart_object` it takes the taxes into account when recalculating the totals (`WC_Cart_Totals` resets the `shipping_tax_total` in the calculation process if the method is not taxable)

If the selected shipping method is set as not taxable, the tax will not be added and the totals will be correct:
screen shot 2018-02-15 at 20 17 36

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -542,12 +542,6 @@ class WC_Taxjar_Integration extends WC_Integration {
 			'line_items' => $line_items,
 		) );
 
-		if ( class_exists( 'WC_Cart_Totals' ) ) { // Woo 3.2+
-			do_action( 'woocommerce_cart_reset', $wc_cart_object, false );
-			do_action( 'woocommerce_before_calculate_totals', $wc_cart_object );
-			new WC_Cart_Totals( $wc_cart_object );
-		}
-
 		foreach ( $this->line_items as $line_item_key => $line_item ) {
 			if ( isset( $cart_taxes[ $this->rate_ids[ $line_item_key ] ] ) ) {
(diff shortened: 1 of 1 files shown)


fastdivision:
Thank you @marcinbot! We'll review this PR shortly.

fastdivision:
@marcinbot Appreciate your help on this. I spent time exploring this PR last week and made some improvements via #61 to fix broken specs. However, moving `new WC_Cart_Totals( $wc_cart_object );` interferes with our ability to fully support product exemptions. See my notes in #61.

Basically, if multiple line items in an order within the same tax class (such as clothing) have different rates based on a threshold ($110 for clothing in New York), it's difficult to rely on WooCommerce to apply the correct tax rates based on the location and tax class alone. We'll likely need to do some more refactoring to support this scenario or continue with things as-is, making sure a merchant's shipping methods are fully taxable.

For now I'm closing this PR and #61 until we come up with a better solution.

PR #61, opened 2018-03-07
Total Calculations Refactor

### LINKED ISSUE #58: Total price at the checkout not calculated correctly if shipping method is not taxable

fastdivision:
This PR fixes #58 and expands on #59 to properly support calculations for partially exempt orders with exempt thresholds sharing the same tax rate. It also fixes broken specs caused by #59 .

In states such as New York, the clothing exemption threshold is $110. Anything under is exempt from sales tax. Given two products with the same `tax_class` shipping to the same location in NY, the tax rate will be set to 8.875% and then 0%. WooCommerce's native calculation process would apply the 0% rate to both products.

Instead, we should only create or update a tax rate if it's greater than 0%:

```php
if ( $line_item->combined_tax_rate ) {
    $this->create_or_update_tax_rate( $line_item_key, $location, $line_item->combined_tax_rate * 100, $tax_class );
}
```

For exempt products under the threshold, temporarily set the tax status to "none":

```php
if ( isset( $this->line_items[ $line_item_key ] ) && ! $this->line_items[ $line_item_key ]->combined_tax_rate ) {
    $product->set_tax_status( 'none' );
}
```

**Specs passing across the board ✨Woo 2.6, 3.0, 3.1, 3.2, 3.3 🎉**

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -547,7 +547,11 @@ class WC_Taxjar_Integration extends WC_Integration {
 			$line_item_key = $product->get_id() . '-' . $cart_item_key;
 
 			if ( isset( $this->line_items[ $line_item_key ] ) && ! $this->line_items[ $line_item_key ]->combined_tax_rate ) {
-				$product->set_tax_status( 'none' );
+				if ( method_exists( $product, 'set_tax_status' ) ) {
+					$product->set_tax_status( 'none' ); // Woo 3.0+
+				} else {
+					$product->set_prop( 'tax_status', 'none' ); // Woo 2.6
+				}
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


fastdivision:
Closing this branch until we devise a better solution to handle partially exempt orders with thresholds.

If two line items exist in an order with the same tax class, with two different rates due to product exemption thresholds (full rates and reduced rates), WooCommerce can't determine which rate belongs to which line item in this refactor. Filters such as `woocommerce_calc_tax` don't provide enough data to determine the current line item in order to support this scenario.

Spec improvements will be pushed for now in #63.

PR #62, opened 2018-03-07
Nexus API Overages

fastdivision:
If a merchant's TaxJar API token becomes invalid or expired, excessive API calls were being made to `/v2/nexus/regions` to sync nexus addresses for each checkout calculation. This PR should fix the problem. For subsequent calculations, the nexus check will be skipped and the calculation request will still go through but it will return back an error.

Once we re-activate the API token in TaxJar, the merchant can manually refresh their nexus addresses from the TaxJar configuration screen inside WP admin.

**Specs passing in 2.6, 3.0, 3.1, 3.2, 3.3 🎉**

### DIFF
--- a/includes/class-wc-taxjar-nexus.php
+++ b/includes/class-wc-taxjar-nexus.php
@@ -11,6 +11,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 class WC_Taxjar_Nexus {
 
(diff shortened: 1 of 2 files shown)


matthewecampbell:
Worked perfectly on the front-end.  Just one call to /v2/verify then one call to /v2/taxes when a token is expired.

Versions tested:

2.6.1, 3.0, 3.1, 3.2, 3.3

PR #63, opened 2018-03-12
Spec Improvements

fastdivision:
General improvements to specs from #61:

- New specs around exempt shipping and product exemption thresholds
- Use `WC()` function similar to Woo's native specs
- Drop `do_action` for `WC()->cart->calculate_totals()`
- Shipping helper class
- Tweak assert precision


fastdivision:
Specs passing in 2.6.x, 3.0, 3.1, 3.2, 3.3

PR #64, opened 2018-03-16
WooCommerce Subscriptions Refactor

fastdivision:
This PR resolves ongoing issues with recurring totals for subscription products between Woo 2.6 and 3.3. The `calculate_tax` method no longer uses class properties that can be modified by cloned carts in WooCommerce Subscriptions for recurring totals. A new method was introduced to calculate tax specifically for recurring totals without affecting the original totals. I also added some helper methods to get address and line item data before calculating tax.

**Versions Tested:**

- [x] Woo 3.3
- [x] Woo 3.2
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -627,6 +627,12 @@ class WC_Taxjar_Integration extends WC_Integration {
 			'shipping_amount' => $woocommerce->shipping->shipping_total,
 			'line_items' => $line_items,
 		) );
+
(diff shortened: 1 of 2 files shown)


matthewecampbell:
Tested and looks good!

✅ 3.33
✅ 3.2
✅ 3.1
✅ 3.0
✅ 2.6

fastdivision:
All specs passing across the board 🎉Let's do this!

PR #70, opened 2018-05-15
Fix compatibility with PHP 5.2 and 5.3

DanReyLop:
We were porting your latest changes back to `woocommerce-services` and we found [these errors](https://travis-ci.org/Automattic/woocommerce-services/builds/378972561) in our CI. The `func_call_that_returns_array()[0]` syntax is only supported in `PHP 5.4` and up.

WooCommerce Services supports `PHP 5.2+` so we already fixed it in our repo.

I'm not sure what PHP version you folks support, seeing some old commits it looks like it's `PHP 5.3`? If so, you should merge this PR :)

Note that I've only fixed this particular error, I didn't test the whole plugin with an old version of PHP.


fastdivision:
Good catch, thanks @DanReyLop! 👏

PR #72, opened 2018-06-06
Backend Tax Calculations for Deleted Items

### LINKED ISSUE #68: Backend tax calculations cause fatal error with deleted items
Heya! When trying to update an order with a product that has been deleted since the order was first placed, a user will encounter an error and won't be able to proceed.

**Steps to Repro**

1. Create an order with a product you can delete later
2. Delete the product permanently
3. Go to the Edit Order page and try to click the 'update' button. See fatal error.

This happens because `wc_get_product( $id )` ([ref](https://docs.woocommerce.com/wc-apidocs/function-wc_get_product.html)) returns `false` in https://github.com/taxjar/taxjar-woocommerce-plugin/blob/9d8e725186776f8c9628ebea59797a8fa41780ec/includes/class-wc-taxjar-integration.php#L617

To solve this, it's possible to use `$unit_price = $item->get_subtotal();` for getting the price and `'taxable' !== $item->get_tax_status()` for getting the tax status.

fastdivision:
Rather than skipping deleted products when performing backend calculations, this PR salvages line item data from the order to include the products in the API request. Resolves #68

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -565,17 +565,13 @@ class WC_Taxjar_Integration extends WC_Integration {
 
 		// Add tax rates manually for Woo 3.0+
 		// Woo 2.6 adds the rates automatically
-		foreach ( $order->get_items() as $item_key => $item ) {
-			if ( is_object( $item ) ) { // Woo 3.0+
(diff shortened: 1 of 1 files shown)


matthewecampbell:
✅ 2.6
✅ 3.0
✅ 3.2.4
✅ 3.3
✅ 3.4.2

Everything looks good here!

Issue #66, opened 2018-03-29
TaxJar calculation causes multiple/redundant shipping calculations to occur


jeffstieler:
Enabling TaxJar calculations (and subsequently WCS taxes) causes multiple shipping calculations to occur when updating cart contents or using the in-cart shipping calculator.

This results in multiple requests to shipping APIs (when using extensions that do so) as well as extra/unnecessary requests to the TaxJar API itself.

This can be seen by using a shipping method extension with a debug mode:

![screen shot 2018-03-29 at 12 59 45-fullpage](https://user-images.githubusercontent.com/63922/38108117-1cd88e72-3351-11e8-8b2f-e029d2617304.png)

Or by inspecting the TaxJar log:

```
03-29-2018 @ 19:04:50 - :::: TaxJar Plugin requested ::::
03-29-2018 @ 19:04:50 - :::: TaxJar API called ::::
03-29-2018 @ 19:04:50 - Received: {"tax":{"order_total_amount":7.65,"shipping":6.65,"taxable_amount":1.0,"amount_to_collect":0.08,"rate":0.0845,"has_nexus":true,"freight_taxable":false,"tax_source":"origin","breakdown":{"taxable_amount":1.0,"tax_collectable":0.08,"combined_tax_rate":0.0845,"state_taxable_amount":1.0,"state_tax_rate":0.047,"state_tax_collectable":0.05,"county_taxable_amount":1.0,"county_tax_rate":0.0185,"county_tax_collectable":0.02,"city_taxable_amount":1.0,"city_tax_rate":0.019,"city_tax_collectable":0.02,"special_district_taxable_amount":0.0,"special_tax_rate":0.0,"special_district_tax_collectable":0.0,"line_items":[{"id":"208-091d584fced301b442654dd8c23b3fc9","taxable_amount":1.0,"tax_collectable":0.08,"combined_tax_rate":0.0845,"state_taxable_amount":1.0,"state_sales_tax_rate":0.047,"state_amount":0.05,"county_taxable_amount":1.0,"county_tax_rate":0.0185,"county_amount":0.02,"city_taxable_amount":1.0,"city_tax_rate":0.019,"city_amount":0.02,"special_district_taxable_amount":0.0,"special_tax_rate":0.0,"special_district_amount":0.0}]}}}
03-29-2018 @ 19:04:50 - :: Tax Rate Found ::
03-29-2018 @ 19:04:50 - Array
(
    [3] => Array
        (
[...]

PR #73, opened 2018-06-06
Perf Improvements (Round 2)

fastdivision:
This PR revisits the shipping package rate purge we added in v1.6 to ensure total calculations are correct when shipping is taxable and a new shipping tax rate is added in Woo 3.0 - 3.1. We're going to drop the change to address performance issues with multiple / redundant shipping calculations mentioned in #66 and seen causing duplicate queries against the database with shipping plugins. Let's test things out again.


fastdivision:
Closing in favor of #108

PR #74, opened 2018-06-07
Skip API request when there are no line items.

jeffstieler:
Fixes https://github.com/Automattic/woocommerce-services/issues/1433.

Skip API request when there aren't any line items - happens during recalculating taxes on an order where all products have been deleted.


fastdivision:
Perfect, thanks for the PR @jeffstieler!

PR #76, opened 2018-06-11
Handle case where there aren’t order line items, but there are shipping charges.

jeffstieler:
From feedback on https://github.com/Automattic/woocommerce-services/pull/1435#issuecomment-396187342:

> The fix stops the invalid API requests. But when I created a product with price set to 0 and then selected a paid shipping options, there were no taxes requested for the shipping price.
> 
> This is because products with falsey price are not added to `$line_items`.

This PR conditionally includes `amount` or `line_items` as referenced in the documentation here: https://developers.taxjar.com/api/reference/#post-calculate-sales-tax-for-an-order

> Either `amount` or `line_items` parameters are required to perform tax calculations.

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -268,7 +268,12 @@ class WC_Taxjar_Integration extends WC_Integration {
 		);
 
 		// Strict conditions to be met before API call can be conducted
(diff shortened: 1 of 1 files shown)


fastdivision:
Looks good to me, thanks @jeffstieler!

PR #77, opened 2018-06-21
Tax Code Precedence for Calculations

fastdivision:
This PR ensures products set to a "None" tax status with a custom product tax class containing a product tax code remain fully exempt. For example:

1) Set `None` tax status for a given product
2) Set `Clothing - 20010` or another custom tax class for the product

Our plugin should send a `99999` tax code for this product, not `20010`.


matthewecampbell:
✅ 2.6
✅ 3.0
✅ 3.2
✅ 3.3
✅ 3.4.3

Everything looks good here with frontend and backend calcs

PR #82, opened 2018-08-09
Call `woocommerce_after_calculate_totals` action for other plugins

### LINKED ISSUE #80: woocommerce_after_calculate_totals action is not performed after totals are recalculated by the plugin
Related issue reported in WooCommerce Services: https://github.com/Automattic/woocommerce-services/issues/1471

The smart coupons plugin is using the `woocommerce_after_calculate_totals` hook to apply its discounts. But the action is not called when the TaxJar plugin recalculates the totals, which results in an erroneous behaviour of that and potentially other plugins.

A potential solution would be to add something like this after the call to [`WC_Cart_Totals`](https://github.com/taxjar/taxjar-woocommerce-plugin/blob/3c2d74fed92cdd0d812333fa07e23554bffd9371/includes/class-wc-taxjar-integration.php#L554):
```
remove_action( 'woocommerce_after_calculate_totals', array( $this, 'calculate_totals' ), 20 );
do_action( 'woocommerce_after_calculate_totals', $wc_cart_object );
add_action( 'woocommerce_after_calculate_totals', array( $this, 'calculate_totals' ), 20 );
```

CC @jeffstieler

fastdivision:
This PR resolves #80 to always call `woocommerce_after_calculate_totals` so other plugins can hook into the process immediately after tax is recalculated in the TaxJar plugin for WooCommerce 3.2+. Since we require discounts and additional data prior to making a sales tax API request and creating a native rate for WooCommerce's internal calculation process, we have to recalculate tax in WooCommerce again after creating the new tax rate and make sure we call the associated actions:

- woocommerce_before_calculate_totals
- woocommerce_after_calculate_totals

Test to make sure general calculations in 3.2+ continue to work. Specs passing.

**Versions Tested:**

- [x] Woo 3.4
- [x] Woo 3.3
- [x] Woo 3.2

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -552,6 +552,9 @@ class WC_Taxjar_Integration extends WC_Integration {
 			do_action( 'woocommerce_cart_reset', $wc_cart_object, false );
 			do_action( 'woocommerce_before_calculate_totals', $wc_cart_object );
 			new WC_Cart_Totals( $wc_cart_object );
+			remove_action( 'woocommerce_after_calculate_totals', array( $this, 'calculate_totals' ), 20 );
+			do_action( 'woocommerce_after_calculate_totals', $wc_cart_object );
+			add_action( 'woocommerce_after_calculate_totals', array( $this, 'calculate_totals' ), 20 );
 		} else {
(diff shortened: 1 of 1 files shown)


matthewecampbell:
✅3.4.4
✅3.3.2
✅3.2.5
✅3.0.0
✅2.6.0

fastdivision:
Automated tests passing across the board 👍

PR #84, opened 2018-08-21
Local Pickup Calculation v2.0 Fix

fastdivision:
This PR fixes local pickup method calculations for v2.0 of our plugin with street address support. When passing the street address, WooCommerce has a hard check for the location to have 4 parts (country, state, postal code, city) when looking up native tax rates. We need to hook into the `woocommerce_matched_rates` filter to support up to 5.

**Versions Tested:**

- [x] Woo 3.4
- [x] Woo 3.3
- [x] Woo 3.2
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6


fastdivision:
Automated tests passing across the board 🎉

PR #85, opened 2018-08-31
 Add filters/actions allowing to override some of the plugin functions

marcinbot:
Over at WooCommerce Services, we are experimenting with integrating our plugins to work better together. Instead of porting the code over to our plugin we decided to try to add hooks that we could use to control the TaxJar plugin.

See https://github.com/Automattic/woocommerce-services/pull/1494 for how the new hooks are being used.

This PR proposes the following hooks:
* `taxjar_should_check_status` allows preventing the connection test request from being sent
* `taxjar_api_token_valid` overrides the API token validation
* `taxjar_can_connect` overrides the connection test result
* `taxjar_download_orders` overrides the value of the download orders option. [Combined with triggering the order validation on saving](https://github.com/Automattic/woocommerce-services/pull/1494/commits/c991c51588c1d40e2cde4412b45ad79b979b2207), this should unlink the account if downloads are enabled and the user opts-in for WooCommerce Services functionality
* `taxjar_method_description` overrides the method description
* `taxjar_enabled` overrides the enabled flag if the user opted in for the taxes functionality. See https://github.com/Automattic/woocommerce-services/pull/1494/commits/531a56e2c7e50d677675cc9c668cb9d847752766
* `taxjar_form_fields` allows modifying the form fields on the settings page
* `taxjar_log` allows extending the logging
* `taxjar_smartcalcs_request` allows overriding the smartcalcs request. If it returns `false`, then the default functionality is used

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -17,7 +17,7 @@ class WC_Taxjar_Integration extends WC_Integration {
 	public function __construct() {
 		$this->id                 = 'taxjar-integration';
 		$this->method_title       = __( 'TaxJar', 'wc-taxjar' );
-		$this->method_description = __( 'TaxJar is the easiest to use sales tax calculation and reporting engine for WooCommerce. Enter your API token (<a href="https://app.taxjar.com/api_sign_up/" target="_blank">click here to get a token</a>), city, and zip code from which your store ships. Enable TaxJar calculations to automatically collect sales tax at checkout. You may also enable order downloads to begin importing transactions from this store into your TaxJar account, all in one click!<br><br><b>For the fastest help, please email <a href="mailto:support@taxjar.com">support@taxjar.com</a>. We\'ll get back to you within hours.</b>', 'wc-taxjar' );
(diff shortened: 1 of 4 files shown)


fastdivision:
Thank you @marcinbot, I'll try to review this PR next week and get these hooks in place for WCS as soon as possible.

includes/class-wc-taxjar-integration.php
```diff
@@ -477,18 +481,20 @@ public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $f
...
-
-		$response = wp_remote_post( $url, array(
-			'headers' => array(
-							'Authorization' => 'Token token="' . $this->settings['api_token'] . '"',
-							'Content-Type' => 'application/json',
-						),
-			'user-agent' => $this->ua,
-			'body' => $json,
-		) );
+		$response = apply_filters( 'taxjar_smartcalcs_request', $json, false );
```
fastdivision: @marcinbot This should return `false` rather than request params from `$json`:

```php
$response = apply_filters( 'taxjar_smartcalcs_request', false, $json );
```

Besides that, LGTM! 👍

marcinbot: Thanks for having a look! I fixed this in 386d9e8

PR #87, opened 2018-10-03
Fix backend order rates after a product tax class is changed

fastdivision:
When modifying an existing order in WP admin or creating a new backend order, line item tax classes remain the same even if the underlying product tax classes change.

If you change a tax class for a product and click the "Recalculate" button without first removing the associated line item, the old tax class will be used for the API request and the new tax class will be used when creating a tax rate from the response. This results in a tax rate created for exempt tax classes greater than 0.

**Specs Passing**

- [x] Woo 3.5
- [x] Woo 3.4
- [x] Woo 3.3
- [x] Woo 3.2
- [x] Woo 3.1
- [x] Woo 3.0
- [x] Woo 2.6


fastdivision:
@dallendalton Is this PR still relevant?

dallendalton:
Nope, it's not relevant anymore. I will go ahead and close it.

PR #93, opened 2018-12-07
Tested up to WooCommerce 3.5

fastdivision:
This PR indicates our plugin is tested up to WooCommerce 3.5 and WordPress 5.0.

**Steps to Reproduce**

1. Upgrade to WordPress 5.0
2. Upgrade to WooCommerce 3.5
3. Perform click-testing

**Expected Result**

All the things work in 3.5 💪

**Click-Test Versions**

- [x] Woo 3.5

**Specs Passing**

- [x] Woo 3.5

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,12 +2,12 @@
 Contributors: taxjar, tonkapark, fastdivision
 Tags: woocommerce, taxjar, tax, taxes, sales tax, tax calculation, sales tax compliance, sales tax filing
 Requires at least: 4.2
-Tested up to: 4.9.2
+Tested up to: 5.0.0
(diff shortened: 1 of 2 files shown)


douglsmith:
As a customer who recently went through the process of selecting TaxJar, I would highly recommend keeping this up to date and in sync with WooCommerce releases—even better if you can update it during the WooCommerce release candidate phase before the final release.

I completely and quickly rejected one of your competitors because they were way behind on WooCommerce compatibility. That was an indicator that they would be always holding us back. Keeping the tested up to string current in your releases says you're on top of things and offers peace of mind to potential and current customers.

PR #100, opened 2019-04-10
Validate zip codes before performing API requests

dallendalton:
The TaxJar plugin works by sending both the store address as well as the ship to address to the TaxJar SmartCalcs API in order to the correct tax rates for each order. This occurs inside of WooCommerce's processes for calculating the order totals. While on the cart (in some themes) or checkout page, each time an end user updates their shipping information WooCommerce will recalculate the totals and cause an additional API request to SmartCalcs. There is already a caching of rates that occurs to help reduce the number of calls to the SmartCalcs API, however rates are only pulled from the cache when there is an exact match to an order (so all the shipping details have to perfectly match). In some themes and situations each character a user enters may cause WooCommerce to recalculate the totals and make a different API request to SmartCalcs. This in turn causes a higher cost to the merchants as use of the SmartCalcs API is billed by the number of requests. Often the request just returns no rates as the zip codes are invalid until the user finishes entering them completely.

This PR solves the issue, and the secondary issue of the user just entering an invalid zip code by validating the zip codes before the API request is performed. The validation method uses the same logic that is present in the SmartCalcs API, so that any request that would have previously returned an error regarding an invalid zip code will skip the request and fall back on the existing rates in WooCommerce (that was the normal functionality after the error was returned from SmartCalcs). This will reduce the number of unnecessary calls to the SmartCalcs API and reduce merchant costs.

**Steps to Reproduce**

1. Enable and configure the TaxJar plugin and enable the logs
2. Add any taxable product to the cart and go to checkout
3. Enter shipping information with an invalid postal code for the country selected
4. Check the logs and you will see that the SmartCalcs API returned an error regarding the invalid zip code. 

**Expected Result**

After applying the PR and performing the above steps, the logs will indicate the the API request was skipped due to an invalid zip code.

Note: For the US, the SmartCalcs API does not except an empty zip code. However, internationally it does. This PR imitates that logic when validating zip codes.

**Click-Test Versions**

- [X] Woo 3.5
- [ ] Woo 3.4
- [ ] Woo 3.3
- [X] Woo 3.2
- [ ] Woo 3.1
- [ ] Woo 3.0
- [X] Woo 2.6

**Specs Passing**

- [X] Woo 3.5
- [X] Woo 3.4
- [x] Woo 3.3
- [X] Woo 3.2
- [x] Woo 3.1
- [X] Woo 3.0
- [X] Woo 2.6

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -985,7 +985,7 @@ class WC_Taxjar_Integration extends WC_Integration {
             'AU' => '/^\d{4}$/',
         );
 
-	    if ( isset( $postal_regexes[$to_country] ) ) {
+	    if ( isset( $postal_regexes[ $to_country ] ) ) {
 
 	        // SmartCalcs api allows requests with no zip codes outside of the US, mark them as valid
 	        if ( empty( $to_zip ) ) {
@@ -996,7 +996,7 @@ class WC_Taxjar_Integration extends WC_Integration {
                 }
             }
 
-	        if ( preg_match( $postal_regexes[$to_country], $to_zip ) === 0 ) {
+	        if ( preg_match( $postal_regexes[ $to_country ], $to_zip ) === 0 ) {
                 $this->_log( ':::: Postal code ' . $to_zip . ' is invalid for country ' . $to_country . ', API request stopped. ::::' );
 	            return false;
             }
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -310,6 +310,12 @@ class WC_Taxjar_Integration extends WC_Integration {
 			return false;
 		}
 
+		// Valid zip codes to prevent unnecessary API requests
+        if ( ! $this->is_postal_code_valid( $to_country, $to_state, $to_zip ) ) {
+            //$this->_log( ':::: Order contains invalid ship to zip code ::::' );
+            return false;
+        }
+
 		$taxjar_nexus = new WC_Taxjar_Nexus( $this );
 
 		if ( ! $taxjar_nexus->has_nexus_check( $to_country, $to_state ) ) {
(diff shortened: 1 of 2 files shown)


includes/class-wc-taxjar-integration.php
```diff
@@ -310,6 +310,12 @@ public function calculate_tax( $options = array() ) {
 			return false;
 		}
 
+		// Valid zip codes to prevent unnecessary API requests
+        if ( ! $this->is_postal_code_valid( $to_country, $to_state, $to_zip ) ) {
+            //$this->_log( ':::: Order contains invalid ship to zip code ::::' );
```
fastdivision: Do we want to keep this log line in the codebase?

includes/class-wc-taxjar-integration.php
```diff
@@ -961,6 +967,45 @@ public function get_taxable_address() {
 		return apply_filters( 'woocommerce_customer_taxable_address', array( $country, $state, $postcode, $city, $street ) );
 	}
 
+	public function is_postal_code_valid( $to_country, $to_state, $to_zip ) {
+
```
fastdivision: Drop newline

includes/class-wc-taxjar-integration.php
```diff
@@ -961,6 +967,45 @@ public function get_taxable_address() {
...
+            'ES' => '/^\d{5}$/',
+            'DK' => '/^\d{4}$/',
+            'SE' => '/^\d{3}[ ]?\d{2}$/',
+            'BE' => '/^\d{4}$/',
+            'IN' => '/^\d{6}$/',
+            'AU' => '/^\d{4}$/',
+        );
+
+	    if ( isset( $postal_regexes[ $to_country ] ) ) {
+
```
fastdivision: Drop newline

includes/class-wc-taxjar-integration.php
```diff
@@ -961,6 +967,45 @@ public function get_taxable_address() {
...
+                } else {
+	                return true;
+                }
+            }
+
+	        if ( preg_match( $postal_regexes[ $to_country ], $to_zip ) === 0 ) {
+                $this->_log( ':::: Postal code ' . $to_zip . ' is invalid for country ' . $to_country . ', API request stopped. ::::' );
+	            return false;
+            }
+
```
fastdivision: Drop newline

tests/specs/test-actions.php
```diff
@@ -1242,4 +1242,132 @@ function test_correct_taxes_for_subscription_products_with_other_products_and_tr
...
+                'P1P 0G0' => false,
+            )
+        );
+
+	    foreach( $postal_array as $country => $codes ) {
+	        foreach( $codes as $code => $expected ) {
+                $this->assertEquals( $this->tj->is_postal_code_valid( $country, null, $code ), $expected );
+            }
+        }
+
```
fastdivision: Drop newline

tests/specs/test-actions.php
```diff
@@ -1242,4 +1242,132 @@ function test_correct_taxes_for_subscription_products_with_other_products_and_tr
 			$this->assertEquals( $recurring_cart->get_taxes_total(), 1.77, '', 0.01 );
 		}
 	}
+
+	function test_is_postal_code_valid() {
+
```
fastdivision: Drop newline

fastdivision:
Click tested locally on WooCommerce 3.5, LGTM!

PR #107, opened 2019-05-01
Recurring Subscription Support

dallendalton:
Before this PR, WooCommerce subscriptions was only partially supported. It worked correctly for a subscription without a trial and for the initial order of a subscription with a trial. However, for the renewal orders from a subscription with a trial, no tax was added. This occurred because the TaxJar tax calculation was never called during the creation of the renewal order and for the case of subscriptions with trials, there was no tax on the original order (because the total was $0) that the renewal order could pull from. This PR solves this problem by triggering the TaxJar tax calculation to occur when the renewal order is created. It also correctly handles the creation of a parent order from a subscription.

Some other changes that affect various issues in the integration:

- Prevent the TaxJar calculate_totals method from running inside of the add to cart action. This just reduces the number of API calls we make to SmartCalcs.
- Validate that the customer object exists in the calculate_tax method. This allows that method to be used when calculating tax from an order object (as is necessary for renewal orders) instead of just from the cart object.

This PR also contains changes to how the specs and testing framework. They are:

- Altering the load order to correctly load the plugins. Previously there were many objects that were instantiated multiple times that not only slowed down the specs but also caused some of the specs to give a false negative in certain circumstances. There were also certain features of the plugin that were not being loaded previously.
- Fix the specs classes so that they do not instantiate their own TaxJar Integration object. This was causing the calculations to run multiple times per spec and it increased as more specs were added.
- Add specs to test the tax calculation on the creation of renewal orders
- Add an additional test class to allow the creation of tests that can perform API calls to the WooCommerce REST API
- Add a spec for testing a customer who is VAT exempt
- Add a subscription helper class to the test framework with a method for creating a subscription through the WooCommerce REST API
- Separate all the subscription specs into their own class and add the functionality to skip the tests if WooCommerce Subscriptions is not installed.

**Steps to Reproduce**

1. Install and configure TaxJar and WooCommerce Subscriptions
2. Create a subscription product that includes a free trial and place an order with it
3. Go into the action scheduler on the WooCommerce status page and run the renewal order action for the subscription just created.
4. Open up the renewal order that was created. You will see it has no tax added.

**Expected Result**

After applying the PR, on step 4 you should that the correct tax has been applied.

**Click-Test Versions**

- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0
- [X] Woo 2.6

**Specs Passing**

- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0
- [X] Woo 2.6

Note: WooCommerce 2.6 is not supported by the most recent version of WooCommerce subscriptions.


includes/class-wc-taxjar-integration.php
```diff
@@ -690,53 +690,66 @@ public function calculate_renewal_order_totals( $order, $subscription, $type ) {
...
 
 	    return $order;
     }
 
+	/**
+	 * Calculate tax on an order
+	 *
+	 * @return null
+	 */
+    public function calculate_order_tax( $order ) {
```
fastdivision: @dallendalton If this method name is only going to be used for WCS renewals, can we make the method name more specific like `calculate_renewal_order_tax`?

includes/class-wc-taxjar-integration.php
```diff
@@ -690,53 +690,66 @@ public function calculate_renewal_order_totals( $order, $subscription, $type ) {
...
+		$this->calculate_order_tax( $subscription );
 
 		$order->calculate_totals();
+		$subscription->calculate_totals();
 
 	    return $order;
     }
 
+	/**
+	 * Calculate tax on an order
```
fastdivision: @dallendalton I'd also be more specific about renewals in the docblock.

includes/class-wc-taxjar-integration.php
dallendalton: @fastdivision So this method will also be used for when we calculate tax on orders created through the API, and any other scenario when we are calculating tax from the order object instead of the cart object.

fastdivision: @dallendalton Understood, thank you!

includes/class-wc-taxjar-integration.php
dallendalton: @fastdivision - We can change it to be more specific, but as I mentioned in my other comment, this will also be used to calculate tax on an order that is created through the API (when that is implemented). Currently this is only used for calculating tax on subscription orders and renewal orders, but in the future it will be used more generally.

Issue #110, opened 2019-05-21
Dynamic tax category filter


ryan-boder:
My company sells a service that TaxJar does not yet have a tax a [tax category](https://developers.taxjar.com/api/reference/#get-list-tax-categories) for. As a workaround we set the product's tax class as "General Services - 19000" and added a few lines to get_line_items() in class-wc-taxjar-integration.php so that in some states TaxJar will treat the product as fully taxable instead of general services. This simulates the tax behavior of our service.

```
// In function get_line_items ...
	if ( isset( $tax_class ) && is_numeric( end( $tax_class ) ) ) {
		$tax_code = end( $tax_class );

		//// Treat general services as regular product in some states
		if ($tax_code === '19000') {
			$address = $this->get_address($wc_cart_object);
			if ($address['to_state'] && in_array($address['to_state'], ['AL', 'AK', 'AZ', 'AR'])) {
				$tax_code = '';
			}
		}
		//////////////////////////////////////////////////////////////////////////////////////////////
	}
```

Would you be willing to add a filter to the plugin so that we can move this code out of the taxjar plugin and put it cleanly in an other plugin or in functions.php? I imagine others might be in a similar situation where the categories taxjar provides aren't sufficient and they need to implement some kind of dynamic tax code logic.

Something like this might work...

```
add_filter('taxjar_get_line_items', function($line_items, $wc_cart_object){
	// Manipulate tax code for each line item
	return $line_items;
});
```

ryan-boder:
Requested PR https://github.com/taxjar/taxjar-woocommerce-plugin/pull/111

dallendalton:
Closed, see comment in https://github.com/taxjar/taxjar-woocommerce-plugin/pull/111 for details of implementation.

PR #111, opened 2019-05-21
Add filter to enable dynamic tax codes

ryan-boder:
Adds the filter requested in [this issue](https://github.com/taxjar/taxjar-woocommerce-plugin/issues/110) so we can get the correct tax behavior for our services without having to modify the taxjar plugin directly.

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -924,7 +924,7 @@ class WC_Taxjar_Integration extends WC_Integration {
 			}
 		}
 
-		return $line_items;
+		return apply_filters( 'taxjar_get_backend_line_items', $line_items, $order, $this );
 	}
 
 	protected function get_line_item( $id, $line_items ) {
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -870,7 +870,7 @@ class WC_Taxjar_Integration extends WC_Integration {
 			}
 		}
 
-		return $line_items;
+		return apply_filters( 'taxjar_get_line_items', $line_items, $wc_cart_object, $this );
 	}
 
 	/**


dallendalton:
@ryan-boder - Thanks for submitting this PR. We are currently evaluating it to see if there are any unexpected issues or problems it would create. One thing has come up as we initially reviewed the changes. 

Ideally we would want one single filter to handle both situations (the cart and back-end order). In order to do this we would probably have to place it outside of those methods, potentially in the calculate_totals method, and instead of passing the cart and order objects we would just use the line items and the address. The reason for wanting only one filter would be ensuring consistency between the various locations. Currently there are only the two places (calculating tax for the cart and for orders created in the back-end), however we are currently working on a real time transaction sync into TaxJar and this would use some of the same logic. Having just one filter would ensure that the exemption code used when calculating tax to collect on an order would match the exemption code when the completed order is sent to TaxJar. Any discrepancies between those would cause our reporting in TaxJar to have inaccuracies.

ryan-boder:
I have it farming the work out to one function anyway. I would definitely prefer it the way you're suggesting but this seems to work as a temporary solution.

```
function update_tax_code_for_alarm_monitoring( $line_items, $to_state ) {
	if ( ! $to_state ) {
		return $line_items;
	}

	return array_map( function ( $line_item ) use ( $to_state ) {
		if ( $line_item['product_tax_code'] === '19000' ) {
			if ( $to_state && in_array( $to_state, [ 'OH', 'AR', 'NJ' ] ) ) {
				$line_item['product_tax_code'] = '';
			}
		}

		return $line_item;
	}, $line_items );
}

add_filter( 'taxjar_get_line_items', function ( $line_items, $wc_cart_object, $taxjar ) {
	$taxable_address = $taxjar->get_taxable_address();
	$to_state        = isset( $taxable_address[1] ) && ! empty( $taxable_address[1] ) ? $taxable_address[1] : false;

	return update_tax_code_for_alarm_monitoring( $line_items, $to_state );
}, 10, 3 );

add_filter( 'taxjar_get_backend_line_items', function ( $line_items, $order, $taxjar ) {
	$to_state = $order->get_shipping_state();

	return update_tax_code_for_alarm_monitoring( $line_items, $to_state );
}, 10, 3 );
```

Of course, it would be even better if TaxJar would add a tax code for alarm monitoring and I didn't have to deal with this but customer service told me there are no plans to do that so I'm stuck hacking in a work around in order to be compliant.

ryan-boder:
@dallendalton Does the upcoming release with real time transaction sync have the single filter you mentioned above that I will be able to use for changing tax codes on a product based on which state the customer is in?

dallendalton:
Hi @ryan-boder. Though ideally we wanted a single filter, we have found it wasn't practical in this situation. However, the upcoming release with Transaction and Customer sync does include filters at those locations you requested. That release is going into a beta that will run for the next couple of weeks. 

If you don't want to participate in the beta, those changes are in https://github.com/taxjar/taxjar-woocommerce-plugin/pull/117 which is scheduled to release on Monday (8-12-19) along with our support for WooCommerce 3.7. 

Just a note that the filters have been slightly altered, mostly just the names to provide a little more clarity as we build additional features in the future. So with just minor adaptations your existing customizations will work correctly. Also, there is a third filter that is included with the Transaction and Customer Sync that your custom code will need to support as well. This filter will allow you to alter the product tax class before a completed transaction is sent to TaxJar. Utilizing this filter with the same logic you used for the others will ensure that the expected tax on reports in TaxJar matches the tax you collected. This filter can be found here https://github.com/taxjar/taxjar-woocommerce-plugin/blob/c18fce3e909fff28ea14f750e0e9e3fe096db384/includes/class-taxjar-order-record.php#L299

Since these filters are now included in separate PRs, I am going to go ahead and close this one. Thanks!

ryan-boder:
@dallendalton I'm trying to use the filters in the 3.0.1 beta but ran into an issue. It looks like the 3rd argument was removed from the 'taxjar_cart_get_line_items' filter. The difference between the calls is

```
// what I submitted
return apply_filters( 'taxjar_get_line_items', $line_items, $wc_cart_object, $this );

// what's in 3.0.1 beta
return apply_filters( 'taxjar_cart_get_line_items', $line_items, $wc_cart_object );
```

but I was using $this (WC_Taxjar_Integration) to call `$taxjar->get_taxable_address()` as shown in the filter implementations in my previous comment above. Is there a way to get the shipping address form $wc_cart_object that I'm missing? Otherwise, how do I get the shipping state to determine whether the item is taxable in that state?

dallendalton:
@ryan-boder - So in order to get the address during the cart calculation you can use the customer object like this:

`$country = WC()->customer->get_shipping_country();
$state = WC()->customer->get_shipping_state();
$postcode = WC()->customer->get_shipping_postcode();
$city = WC()->customer->get_shipping_city();
$street = WC()->customer->get_shipping_address();`

Using this inside of the taxjar_cart_get_line_items gets the current customer data that is associated with the cart. However, we do have some logic in our get_taxable_address method that can alter these values to handle situations like local pickup shipping methods. Because of this we have decided that we need to add back in the WC_Taxjar_Integration object as a parameter like it was in the version you submitted. This will be updated in the repository today. It will also be included in the general release of the transaction and customer sync that will occur later this week (probably tomorrow). Thanks!

ryan-boder:
Thanks!

PR #121, opened 2019-08-06
Tax Rate Lookup State Containing Space Fix

dallendalton:
When placing orders outside of the US, there was an issue that occurred when the state field (can be called other names depending on the country) contained a space. An example of this is "West Sussex". When this is entered at checkout, the plugin would perform the API request to get the rates and insert them into WooCommerce's tax rate table. However when during the calculations, and we attempt to look up the rate in the table it would never be found. This is due to WooCommerce sanitizing the state field with the sanitize_key function when inserting the rate (which removes the space), but neglecting to do this during the rate lookup. This PR fixes the issue by performing the same sanitation before doing the lookup in the affected WooCommerce versions.

**Steps to Reproduce**

1. Add and item to cart and go to checkout
2. Enter an international address with the state field having a space in it.
3. Tax will not be calculated on the order as the rate saved during the API request will not be found

**Expected Result**

After applying the PR, on step 3 tax will be correctly calculated for the order.

**Click-Test Versions**

- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

**Specs Passing**

- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -459,13 +459,24 @@ class WC_Taxjar_Integration extends WC_Integration {
 			'tax_rate_class' => $tax_class,
 		);
 
-		$wc_rate = WC_Tax::find_rates( array(
-			'country' => $location['to_country'],
-			'state' => sanitize_key( $location['to_state'] ),
-			'postcode' => $location['to_zip'],
-			'city' => $location['to_city'],
-			'tax_class' => $tax_class,
-		) );
+		if ( version_compare( WC()->version, '3.2.0', '<' ) ) {
+			$wc_rate = WC_Tax::find_rates( array(
+				'country' => $location['to_country'],
+				'state' => $location['to_state'],
+				'postcode' => $location['to_zip'],
+				'city' => $location['to_city'],
+				'tax_class' => $tax_class,
+			) );
+        } else {
+			$wc_rate = WC_Tax::find_rates( array(
+				'country' => $location['to_country'],
+				'state' => sanitize_key( $location['to_state'] ),
+				'postcode' => $location['to_zip'],
+				'city' => $location['to_city'],
+				'tax_class' => $tax_class,
+			) );
+        }
+
 
 		if ( ! empty( $wc_rate ) ) {
 			$this->_log( ':: Tax Rate Found ::' );
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -459,25 +459,19 @@ class WC_Taxjar_Integration extends WC_Integration {
 			'tax_rate_class' => $tax_class,
 		);
 
-		if ( version_compare( WC()->version, '3.2.0', '<' ) ) {
-			$wc_rate = WC_Tax::find_rates( array(
-				'country' => $location['to_country'],
-				'state' => $location['to_state'],
-				'postcode' => $location['to_zip'],
-				'city' => $location['to_city'],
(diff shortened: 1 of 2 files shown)


includes/class-wc-taxjar-integration.php
```diff
@@ -459,13 +459,24 @@ public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $f
...
+        } else {
+			$wc_rate = WC_Tax::find_rates( array(
+				'country' => $location['to_country'],
+				'state' => sanitize_key( $location['to_state'] ),
+				'postcode' => $location['to_zip'],
+				'city' => $location['to_city'],
+				'tax_class' => $tax_class,
+			) );
+        }
+
```
fastdivision: Remove additional newline

includes/class-wc-taxjar-integration.php
```diff
@@ -459,13 +459,24 @@ public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $f
...
+				'country' => $location['to_country'],
+				'state' => $location['to_state'],
+				'postcode' => $location['to_zip'],
+				'city' => $location['to_city'],
+				'tax_class' => $tax_class,
+			) );
+        } else {
+			$wc_rate = WC_Tax::find_rates( array(
+				'country' => $location['to_country'],
+				'state' => sanitize_key( $location['to_state'] ),
```
fastdivision: If `sanitize_key` is the only difference here, can we DRY up this code?

includes/class-wc-taxjar-integration.php
dallendalton: Resolved in most recent commit

includes/class-wc-taxjar-integration.php
dallendalton: Resolved in most recent commit

includes/class-wc-taxjar-integration.php
```diff
@@ -459,14 +459,19 @@ public function create_or_update_tax_rate( $location, $rate, $tax_class = '', $f
...
 			'city' => $location['to_city'],
 			'tax_class' => $tax_class,
-		) );
+		);
 
+		if ( version_compare( WC()->version, '3.2.0', '>=' ) ) {
+			$rate_lookup['state'] = sanitize_key( $location['to_state'] );
+		}
+		
+		$wc_rate = WC_Tax::find_rates( $rate_lookup );
```
fastdivision: Newline below 474

allendav:
From my testing, this approach does help ensure the entry is created or updated by the tax jar plugin, but it seems WooCommerce core itself, when fetching the line item tax is unable to find the entry.

I think this fix and a fix in WooCommerce core might be needed? I've opened the issue here: https://github.com/woocommerce/woocommerce/issues/24354

dallendalton:
@allendav - You are correct that this fix only applies to stores that use the TaxJar plugin. It was necessary for us to make a fix for our plugin because it appears this issue was introduced in WooCommerce 3.2.0 and we want to maintain backward compatibility back to 3.0.0. I also agree with you that a fix is needed in the WooCommerce core and I believe there are issues opened now in WooCommerce and the WooCommerce services plugin so hopefully they will resolve it in the core relatively soon. Thanks!

PR #133, opened 2019-10-02
Sync Queue Improvements

dallendalton:
The main issue that this PR resolves is fixing records that get stuck in awaiting status in the sync queue. This was happening due to a validation on the batch ID that caused it to not be able to be updated to 0 after a record failed to sync. Since the batch ID remained set to the initial batch ID that processed it, it would never get picked up and placed into another batch and so would stay stuck in the awaiting status.

This PR fixes the issue by altering the validation to allow the batch ID to be reset to 0. It also includes an update function that will run once upon updating and will located all of the records that meet this stuck condition and resolve their batch IDs to 0, allowing them to continue in the sync process.

This PR also includes various other improvements, such as:

- Adding the last error to the sync queue view to allow customers to immediately determine the cause of orders failing to sync.
- Improved error messaging.
- Displaying the batch ID in the sync queue view (can help in troubleshooting issues).
- Additional error handling in sync process.
- Ensuring sync date is set when orders are manually synced.
- Clearing rates from table when they no longer have nexus and nexus is updated.

**Steps to Reproduce**

1. Place an order with the shipping address zip code that is not in the shipping state.
2. Wait for order to sync through normal sync process.
3. Order will attempt to sync and fail, but will stay in "awaiting" status. 
4. Order will have retry attempts of 1 but will never attempt to sync again.

**Expected Result**

After applying the PR, order will retry syncing up to 3 times and then will be placed into failed status.

**Click-Test Versions**

- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

**Specs Passing**

- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# 3.0.10 (2019-10-2)
+* Fix record stuck in awaiting status in sync queue
+* Display last sync error in sync queue
+* Clear regions not in nexus from rate table when nexus is updated
+* Improve error messaging in logs
+* Set synced date on orders when sync is manually triggered
+* Display batch ID in sync queue table
+* Handle unexpected exemptions during sync
+
 # 3.0.9 (2019-09-18)
 * Update validation to support new TaxJar product categories
 * Fix missing filter on refund reference IDs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 3.0.10 (2019-10-2)
+# 3.0.10 (2019-10-04)
 * Fix record stuck in awaiting status in sync queue
 * Display last sync error in sync queue
 * Clear regions not in nexus from rate table when nexus is updated
--- a/includes/abstract-class-taxjar-record.php
+++ b/includes/abstract-class-taxjar-record.php
@@ -214,7 +214,7 @@ abstract class TaxJar_Record {
 			}
 
 			if ( is_wp_error( $response ) ) {
-				$this->sync_failure( __( 'WP_Error occurred on ' . $last_request . ' request - ' , 'wc-taxjar' ) . $response->get_error_message() );
+				$this->sync_failure( __( 'WP_Error occurred on ' . $last_request . ' request. Details: ' , 'wc-taxjar' ) . $response->get_error_message() );
 				return false;
 			}
 
@@ -224,7 +224,36 @@ abstract class TaxJar_Record {
 			}
 
 			if ( in_array( $response[ 'response' ][ 'code' ], $error_responses ) ) {
-				$this->sync_failure( __(  ucfirst( $last_request ) . ' request failed with code: ' , 'wc-taxjar' ) . $response[ 'response' ][ 'code' ] . ' Request: ' . $this->get_last_request() . ' Response: ' . $response[ 'body' ] );
+
+				switch( $response[ 'response' ][ 'code' ] ) {
+					case 400:
+						if ( ! empty( $response['body'] ) ) {
+							$error_message = "Invalid request sent to TaxJar. Details: ";
+							$body = json_decode( $response['body'] );
+
+							if ( ! empty( $body->detail ) ) {
+								$error_message .= $body->detail;
+							}
+						}
+						break;
+					case 401:
+						$error_message = "Authorization error. Please check API key.";
+						break;
+					case 403:
+						$error_message = "Authorization error. Please check API key.";
+						break;
+					case 404:
+						$error_message = "Record does not exist in TaxJar, could not update.";
+						break;
+					case 429:
+						$error_message = "Rate limit reached.";
+						break;
+					default:
+						$error_message = "Error in request, TaxJar response code " . $response[ 'response' ][ 'code' ];
+				}
+
+				$this->sync_failure( $error_message );
+				$this->log( ' Request: ' . $this->get_last_request() . ' Response: ' . $response[ 'body' ] );
 				return false;
 			}
 
--- a/includes/abstract-class-taxjar-record.php
+++ b/includes/abstract-class-taxjar-record.php
@@ -224,7 +224,6 @@ abstract class TaxJar_Record {
 			}
 
 			if ( in_array( $response[ 'response' ][ 'code' ], $error_responses ) ) {
-
 				switch( $response[ 'response' ][ 'code' ] ) {
 					case 400:
 						if ( ! empty( $response['body'] ) ) {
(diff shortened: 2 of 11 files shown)


includes/abstract-class-taxjar-record.php
```diff
@@ -79,14 +81,22 @@ public function create() {
...
+			'batch_id'           => $this->get_batch_id(),
+			'created_datetime'   => $this->get_created_datetime(),
+			'force_push'         => $this->get_force_push()
 		);
 
+		if ( ! empty( $this->get_processed_datetime() ) ) {
+			$insert[ 'processed_datetime' ] = $this->get_processed_datetime();
+		}
+
+		if ( $this->get_last_error() === "" | ! empty( $this->get_last_error() ) ) {
```
fastdivision: @dallendalton Should this be `||` instead of `|`?

includes/abstract-class-taxjar-record.php
```diff
@@ -121,14 +131,18 @@ public function save() {
...
-		if ( ! empty( $this->get_batch_id() ) ) {
+		if ( $this->get_batch_id() === 0 || $this->get_batch_id() === '0' || ! empty( $this->get_batch_id() ) ) {
 			$data[ 'batch_id' ] =  $this->get_batch_id();
 		}
 
 		if ( ! empty( $this->get_retry_count() ) ) {
 			$data[ 'retry_count' ] =  $this->get_retry_count();
 		}
 
+		if ( $this->get_last_error() === "" | ! empty( $this->get_last_error() ) ) {
```
fastdivision: @dallendalton Same thing here for `|`

includes/abstract-class-taxjar-record.php
```diff
@@ -155,76 +169,107 @@ public function get_data() {
...
 			}
-		}
 
-		if ( is_wp_error( $response ) ) {
-			$this->sync_failure();
-			$this->log( __( 'WP_Error occurred on ' . $last_request . ' request - ' , 'wc-taxjar' ) . $response->get_error_message() );
-			return false;
-		}
+			if ( in_array( $response[ 'response' ][ 'code' ], $error_responses ) ) {
+
```
fastdivision: Drop newline

readme.txt
```diff
@@ -91,6 +91,15 @@ Yes. The fee is $19.95 per state, per filing.
 
 == Changelog ==
 
+= 3.0.10 (2019-10-2)
```
fastdivision: Let's use `2019-10-02` for consistency

includes/class-wc-taxjar-queue-list.php
```diff
@@ -174,6 +180,8 @@ public function get_columns() {
 			'created_datetime'      => __( 'Created Time', 'wc-taxjar' ),
 			'processed_datetime'    => __( 'Sync Time', 'wc-taxjar' ),
 			'retry_count'           => __( 'Retry Count', 'wc-taxjar' ),
+			'batch_id'              => __( 'Batch ID', 'wc-taxjar' ),
+			'last_error'              => __( 'Last Error', 'wc-taxjar' ),
```
fastdivision: Minor spacing / alignment issue here

includes/class-wc-taxjar-nexus.php
```diff
@@ -142,4 +143,25 @@ public function get_nexus_from_cache() {
...
+	/**
+	 * Clear non US nexus states from rates table to prevent tax calculation when nexus is removed in a state
+	 * @param $regions - array of nexus regions from API request
+	 */
+	public function clear_non_nexus_rates( $regions ) {
+		global $wpdb;
+		$nexus_states = array();
+
+		foreach( $regions as $region ) {
+			if ( ! empty( $region->country_code ) && $region->country_code === 'US' ){
```
fastdivision: Space between `)` and `{`

includes/abstract-class-taxjar-record.php
dallendalton: Yeah it should have.... turns out php also accepts the single | operator with no issues, though I don't see it documented. Which is why none of the specs or my manual testing had issues.

includes/class-wc-taxjar-nexus.php
```diff
@@ -142,4 +143,25 @@ public function get_nexus_from_cache() {
...
+				if ( ! empty( $region->region_code ) ) {
+					$nexus_states[] = $region->region_code;
+				}
+			}
+		}
+
+		$nexus_states_string = join( "','", $nexus_states );
+		$query = "DELETE FROM {$wpdb->prefix}woocommerce_tax_rates WHERE tax_rate_country = 'US' AND tax_rate_state NOT IN ('{$nexus_states_string}')";
```
fastdivision: So we purge tax rates after installing and back them up. Have we seen or anticipated a scenario where a customer may want to enter a manual tax rate for a non-nexus US state after installation, such as using a flat rate instead of having TaxJar calculate? We could update the WHERE clause to check for `%s Tax` format in the `tax_rate_name`, but it may be overcautious.

includes/class-wc-taxjar-install.php
```diff
@@ -133,6 +141,7 @@ private static function get_schema() {
   created_datetime datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
   processed_datetime datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
   retry_count smallint(4) NOT NULL DEFAULT 0,
+  last_error longtext NOT NULL DEFAULT '',
```
fastdivision: Do we anticipate the need for `longtext` here or would `text` be sufficient?

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# 3.0.10 (2019-10-2)
```
fastdivision: Tweak `2019-10-2` to `2019-10-02` for consistency

includes/class-wc-taxjar-nexus.php
dallendalton: I haven't seen or heard of that scenario before, though it could be possible. I feel fine with the code as is right now because I can't think of a scenario where someone would want to take that approach. Plus I think we want to discourage people from altering the rates in that way for a US location. They are basically ensuring they will not have the most accurate rates.

fastdivision: @dallendalton Sounds good, resolving this 👍

includes/class-wc-taxjar-install.php
dallendalton: You are right, text will be plenty sufficient for this column.

PR #139, opened 2019-12-19
Disable Date Validation Filters

dallendalton:
Currently the plugin only supports syncing orders that have completed or refunded status. Previously there were filters added to enable merchants to sync other statuses. However there is also a completed date validation that prevents orders that have never reached completed status from syncing. This prevented merchants from syncing orders with pending or other non complete statuses. This PR adds additional filters to allow merchants to bypass the date validation in these cases.

**Click-Test Versions**

- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

**Specs Passing**

- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

### DIFF
--- a/includes/class-taxjar-refund-record.php
+++ b/includes/class-taxjar-refund-record.php
@@ -27,9 +27,11 @@ class TaxJar_Refund_Record extends TaxJar_Record {
 			return false;
 		}
 
-		if ( empty( $this->order_completed_date ) ) {
-			$this->add_error( __( 'Refund failed validation - parent order has no completed date.', 'wc-taxjar' ) );
-			return false;
+		if ( apply_filters( 'taxjar_should_validate_order_completed_date', true ) ) {
+			if ( empty( $this->order_completed_date ) ) {
+				$this->add_error( __( 'Refund failed validation - parent order has no completed date.', 'wc-taxjar' ) );
+				return false;
+			}
 		}
 
 		$valid_order_statuses = apply_filters( 'taxjar_valid_order_statuses_for_sync', array( 'completed', 'refunded' ) );
--- a/includes/class-taxjar-refund-record.php
+++ b/includes/class-taxjar-refund-record.php
@@ -27,7 +27,7 @@ class TaxJar_Refund_Record extends TaxJar_Record {
 			return false;
 		}
 
-		if ( apply_filters( 'taxjar_should_validate_order_completed_date', true ) ) {
+		if ( WC_Taxjar_Transaction_Sync::should_validate_order_completed_date() ) {
 			if ( empty( $this->order_completed_date ) ) {
 				$this->add_error( __( 'Refund failed validation - parent order has no completed date.', 'wc-taxjar' ) );
 				return false;
--- a/includes/class-wc-taxjar-transaction-sync.php
+++ b/includes/class-wc-taxjar-transaction-sync.php
@@ -70,7 +70,7 @@ class WC_Taxjar_Transaction_Sync {
 			return $actions;
 		}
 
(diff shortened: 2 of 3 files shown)


includes/class-taxjar-refund-record.php
```diff
@@ -27,9 +27,11 @@ public function should_sync() {
 			return false;
 		}
 
-		if ( empty( $this->order_completed_date ) ) {
-			$this->add_error( __( 'Refund failed validation - parent order has no completed date.', 'wc-taxjar' ) );
-			return false;
+		if ( apply_filters( 'taxjar_should_validate_order_completed_date', true ) ) {
```
codymullins: From the sounds of the issue and change, I would consider inverting this condition. This makes the validation of order completed date "opt-in" if the value is not set. 

If we instead add a filter for _bypassing_ the validation here it would feel safer since I would imagine we would want this validation in most cases. (e.g. - `bypass_order_completed_validation`).

dallendalton: We talked about this in another place but putting here as well. The filter does default to true and will only return false if someone adds a custom filter to it in a plugin or in their funtions.php file. Also, the latest commit does make the intended functionality a little easier to read.

includes/class-wc-taxjar-transaction-sync.php
```diff
@@ -70,10 +70,13 @@ public function add_order_meta_box_action( $actions ) {
...
 
-		if ( empty( $theorder->get_date_completed() ) ) {
-			return $actions;
+		if ( WC_Taxjar_Transaction_Sync::should_validate_order_completed_date() ) {
+			if ( empty( $theorder->get_date_completed() ) ) {
+				return $actions;
+			}
 		}
 
+
```
fastdivision: Extra newline here

includes/class-wc-taxjar-transaction-sync.php
```diff
@@ -620,40 +625,41 @@ public function get_orders_to_backfill( $start_date = null, $end_date = null, $f
...
-				AND p.post_date < '{$end_date}' 
-				AND order_meta_completed_date.meta_value IS NOT NULL 
-				AND order_meta_completed_date.meta_value != '' 
-				ORDER BY p.post_date ASC
-				", ARRAY_N
-			);
+			$query = "SELECT p.id FROM {$wpdb->posts} AS p ";
+
+			if ( $should_validate_completed_date ) {
+				$query .= "INNER JOIN {$wpdb->postmeta} AS order_meta_completed_date ON ( p.id = order_meta_completed_date.post_id )  AND ( order_meta_completed_date.meta_key = '_completed_date' ) ";
```
fastdivision: Seeing an additional space here on lines 634, 648, 651 between `)` and `AND`: `)  AND (`

includes/class-wc-taxjar-transaction-sync.php
dallendalton: Fixed in most recent commit, thanks!

includes/class-wc-taxjar-transaction-sync.php
dallendalton: Good catch, fixed in the most recent commit.

PR #143, opened 2020-01-27
Action scheduler 3.0 fix

dallendalton:
Action Scheduler 3.0 changes how scheduled actions are stored in the database. Instead of being a custom post they are now stored in their own table. As part of this change they have limit the arguments field to 190 characters. Previously this was a longtext field used for storing a JSON string of the action arguments. With this change, TaxJar transaction sync batches with lots of orders would fail to migrate since their arguments JSON string was over the character limit. This would cause those batches to never run and sync to TaxJar.

In order to fix this we have changed our approach to using Action Scheduler. Instead of processing the queue and then running the transactions in batches in a later scheduled action, we have simplified our approach to process the queue and immediately begin syncing a set number of records from the queue (the batch size). Then if more transactions remain to be synced it will trigger the queue to process again at the next possible moment. Otherwise it will process the queue again at the normal schedule (every 20 minutes). 

This both makes the plugin fully compatible with Action Scheduler 3.0. It also resolves a previous issue where when batches timed out they would never retry and those orders in the batch would never be synced. (Now if a batch fails, the transactions will be resynced on the next queue process action).

**Steps to Reproduce**

1. Create or update 50 records so the sync queue has a least 50 records awaiting to be synced.
2. Process the queue so that there is a full batch
3. Install a plugin that uses action scheduler 3.0 and let the migration run
4. There will be a notice that the batch action failed to migrate

**Expected Result**

After applying the PR, this should happen.

Since we are no longer syncing in batches, when this version is installed it will remove all the batch scheduled actions and will process the queue every 20 minutes. So following the steps above the notification will not be given and then the transactions will sync normally.

**Click-Test Versions**

- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

**Specs Passing**

- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

### DIFF
--- a/includes/class-wc-taxjar-record-queue.php
+++ b/includes/class-wc-taxjar-record-queue.php
@@ -22,23 +22,6 @@ class WC_Taxjar_Record_Queue {
 	}
 
 	/**
-	 * Update record status and batch id in queue
-	 *
-	 * @param array $queue_ids - array of queue ids that were added to batch
-	 * @param array $batch_id - id of batch
-	 * @return null
-	 */
-	static function add_records_to_batch( $queue_ids, $batch_id ) {
-		global $wpdb;
-
-		$table_name = self::get_queue_table_name();
-		$queue_ids_string = join( "','", $queue_ids );
-
-		$query = "UPDATE {$table_name} SET batch_id = {$batch_id} WHERE queue_id IN ('{$queue_ids_string}')";
-		$wpdb->get_results( $query );
-	}
-
-	/**
 	 * Get the queue ids of all active (new and awaiting) records in queue that are not in a batch
 	 *
 	 * @return array|bool - if active records are found returns array, otherwise returns false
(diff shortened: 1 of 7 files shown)


includes/class-wc-taxjar-record-queue.php
```diff
@@ -22,38 +22,54 @@ static function get_queue_table_name() {
...
+	static function get_all_active_in_queue() {
 		global $wpdb;
 
 		$table_name = self::get_queue_table_name();
-		$queue_ids_string = join( "','", $queue_ids );
 
-		$query = "UPDATE {$table_name} SET batch_id = {$batch_id} WHERE queue_id IN ('{$queue_ids_string}')";
-		$wpdb->get_results( $query );
+		$query = "SELECT queue_id FROM {$table_name} WHERE status IN ( 'new', 'awaiting' )";
+		$results = $wpdb->get_results( $query,  ARRAY_A );
```
fastdivision: @dallendalton Extra space after comma: `,  ARRAY_A`

includes/class-wc-taxjar-record-queue.php
```diff
@@ -22,38 +22,54 @@ static function get_queue_table_name() {
...
 	 */
-	static function get_all_active_in_queue() {
+	static function get_active_records_to_process( $number_to_process ) {
 		global $wpdb;
 
 		$table_name = self::get_queue_table_name();
 
-		$query = "SELECT queue_id FROM {$table_name} WHERE status IN ( 'new', 'awaiting' ) AND batch_id = 0";
+		$query = "SELECT * FROM {$table_name} WHERE status IN ( 'new', 'awaiting' ) LIMIT {$number_to_process}";
 		$results = $wpdb->get_results( $query,  ARRAY_A );
```
fastdivision: @dallendalton Extra space after comma: `,  ARRAY_A`

PR #145, opened 2020-03-10
Improve Security of Ajax Requests

dallendalton:
Currently we have two ajax requests, to update nexus and to trigger a transaction backfill. Right now, neither the page that called the request or the permissons of the user are been validated. This means that there is the potential for anyone to be able to call these methods from outside of the site without having proper permissions.

This PR resolves the issue be creating security nonces that are loaded on the TaxJar settings page. These nonces are then send with the ajax requests and validated. This ensures that the ajax methods can only be called from those pages. This PR also validates the the current user has edit order permissions before the rest of the method runs.

This PR also includes a minor change to the load order when setting up the unit tests (necessary for testing on WooCommerce 4.0) as well as an update to the PR template to account for newer versions of WooCommerce.

**Steps to Reproduce**

1. Create a user with edit orders permissions
2. Attempt to sync nexus
3. Check the logs and confirm nexus was synced

**Expected Result**

After applying the PR, an error message will be displayed when attempting the request from an invalid page or when the user doesn't have the necessary permissions.

**Click-Test Versions**

- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

**Specs Passing**

- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2
- [X] Woo 3.1
- [X] Woo 3.0

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,20 +12,28 @@ After applying the PR, this should happen.
 
 **Click-Test Versions**
 
+- [ ] Woo 4.0
+- [ ] Woo 3.9
+- [ ] Woo 3.8
+- [ ] Woo 3.7
+- [ ] Woo 3.6
 - [ ] Woo 3.5
 - [ ] Woo 3.4
 - [ ] Woo 3.3
 - [ ] Woo 3.2
 - [ ] Woo 3.1
 - [ ] Woo 3.0
-- [ ] Woo 2.6
 
 **Specs Passing**
 
+- [ ] Woo 4.0
+- [ ] Woo 3.9
+- [ ] Woo 3.8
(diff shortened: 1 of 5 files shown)


n9yty:
Now that this is public, having an expedited patch would be appreciated. ;) Thanks for the hard work!

PR #146, opened 2020-03-12
If we don't have any $taxes, bail early

barryhughes:
Please see the [note in the community forums](https://wordpress.org/support/topic/calculate_tax-return-type/): we're getting unwanted 'noise' in the error logs, because a potentially false value is treated like an array. If it is not an array, there are no taxes to work with so we can just bail out early :-)


dallendalton:
This is being taken care of by https://github.com/taxjar/taxjar-woocommerce-plugin/pull/148. It will be included in our next release.

PR #163, opened 2020-07-06
INT-1043 Shipping Rate Fix

dallendalton:
In certain scenarios, when a cart has a reduced rate tax code and shipping is fully taxable, the shipping may not get taxed at the correct rate. This PR resolves this issue.

Also, Colorado was recently changed in TaxJar to have shipping not be taxable by default. This change broke some of the integration tests that tested the shipping tax on orders to Colorado. This PR fixes those integration tests.

**Steps to Reproduce**

The test_taxes_with_different_shipping_and_item_rates integration test simulates the scenario where the issue was occurring. Adding and then running this test in an older version of the integration will cause the test to fail (this can also be done on the front end of a store by using the address, store address and product data contained in the integration test).

**Expected Result**

After apply this PR that test will pass successfully.

**Click-Test Versions**

- [X] Woo 4.2
- [X] Woo 4.1
- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2

**Specs Passing**

- [X] Woo 4.2
- [X] Woo 4.1
- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -638,9 +638,14 @@ class WC_Taxjar_Integration extends WC_Settings_API {
 			$taxes['freight_taxable']    = (int) $taxjar_response->freight_taxable;
 			$taxes['has_nexus']          = (int) $taxjar_response->has_nexus;
 			$taxes['tax_rate']           = $taxjar_response->rate;
-			$taxes['shipping_rate']      = $taxjar_response->breakdown->shipping->combined_tax_rate;
+			$taxes['shipping_rate']      = $taxjar_response->rate;
 
 			if ( ! empty( $taxjar_response->breakdown ) ) {
+
+			    if ( ! empty( $taxjar_response->breakdown->shipping ) ) {
+				    $taxes['shipping_rate'] = $taxjar_response->breakdown->shipping->combined_tax_rate;
+                }
+
 				if ( ! empty( $taxjar_response->breakdown->line_items ) ) {
 					$line_items = array();
 					foreach ( $taxjar_response->breakdown->line_items as $line_item ) {
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -638,6 +638,7 @@ class WC_Taxjar_Integration extends WC_Settings_API {
 			$taxes['freight_taxable']    = (int) $taxjar_response->freight_taxable;
 			$taxes['has_nexus']          = (int) $taxjar_response->has_nexus;
 			$taxes['tax_rate']           = $taxjar_response->rate;
+			$taxes['shipping_rate']      = $taxjar_response->breakdown->shipping->combined_tax_rate;
 
 			if ( ! empty( $taxjar_response->breakdown ) ) {
 				if ( ! empty( $taxjar_response->breakdown->line_items ) ) {
@@ -689,7 +690,7 @@ class WC_Taxjar_Integration extends WC_Settings_API {
 			// Add shipping tax rate
 			$taxes['rate_ids']['shipping'] = $this->create_or_update_tax_rate(
 				$location,
-				$taxes['tax_rate'] * 100,
+				$taxes['shipping_rate'] * 100,
 				'',
 				$taxes['freight_taxable']
 			);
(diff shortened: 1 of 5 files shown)


includes/class-wc-taxjar-integration.php
```diff
@@ -638,8 +638,14 @@ public function calculate_tax( $options = array() ) {
 			$taxes['freight_taxable']    = (int) $taxjar_response->freight_taxable;
 			$taxes['has_nexus']          = (int) $taxjar_response->has_nexus;
 			$taxes['tax_rate']           = $taxjar_response->rate;
```
fastdivision, line 640: @dallendalton Do we still use `$taxes['tax_rate']` anywhere after this change?

includes/class-wc-taxjar-integration.php
```diff
@@ -825,6 +831,7 @@ public function calculate_totals( $wc_cart_object ) {
 
 		$address = $this->get_address( $wc_cart_object );
 		$line_items = $this->get_line_items( $wc_cart_object );
+		$shipping_total = $wc_cart_object->get_shipping_total();
```
fastdivision, line 834: @dallendalton Just changing this to use the `get_shipping_total` helper method instead?

tests/specs/test-actions.php
```diff
@@ -45,29 +46,77 @@ function test_correct_taxes_with_shipping() {
...
-			$this->assertEquals( $item['line_tax'], 0.73, '', 0.01 );
+			$this->assertEquals( $item['line_tax'], 0.89, '', 0.01 );
 		}
 
 		WC()->session->set( 'chosen_shipping_methods', array() );
 		TaxJar_Shipping_Helper::delete_simple_flat_rate();
 	}
 
+	function test_taxes_with_different_shipping_and_item_rates() {
+		update_option( 'woocommerce_tax_classes', "Reduced rate\nZero Rate\nnon-prescription-51010" );
```
fastdivision, line 79: Consider capitalizing `non-prescription-51010`, blends in with the escape sequence `\n`.

includes/class-wc-taxjar-integration.php
```diff
@@ -689,7 +695,7 @@ public function calculate_tax( $options = array() ) {
 			// Add shipping tax rate
 			$taxes['rate_ids']['shipping'] = $this->create_or_update_tax_rate(
 				$location,
-				$taxes['tax_rate'] * 100,
+				$taxes['shipping_rate'] * 100,
 				'',
 				$taxes['freight_taxable']
```
ScottRudiger, line 700: Missing trailing comma, for consistency w/ other lines in file

tests/specs/test-actions.php
```diff
@@ -45,29 +46,77 @@ function test_correct_taxes_with_shipping() {
...
+			'state' => 'IL',
+			'zip' => '62294',
+			'city' => 'Troy',
+		) );
+
+		$product_options = array(
+			'tax_class' => 'non-prescription-51010',
+			'name'          => 'Test Prescription',
+			'price'         => 11.99,
+			'sku'           => 'TEST-PRES'
```
ScottRudiger, line 103: Trailing comma,

includes/class-wc-taxjar-integration.php
dallendalton: Yeah we had to do it the other way for older versions of WooCommerce we no longer support. During all future changes I will be making updates to use the proper ways.

includes/class-wc-taxjar-integration.php
dallendalton: The correct way is actually not to have a trailing comma. All these formatting/styling inconsistencies will be fixed in a future PR. Due to the number of them I want to do a standalone PR for them so that any functional changes do not get lost in the PR.

tests/specs/test-actions.php
dallendalton: Same as above. Thanks!

includes/class-wc-taxjar-integration.php
dallendalton: Nope we don't, I have removed it. Thanks!

tests/specs/test-actions.php
dallendalton: Good idea, it is done. Thanks!

PR #165, opened 2020-08-25
Fix international from address validation

dallendalton:
There is validation is place in order to prevent attempting to sync transactions to TaxJar, when the transactions have known missing or bad data. One of those validations is for the ship from address and it confirms that there is a country, state, city and zip code present before attempting to sync the transaction. This causes issues with international address that do not have states or provinces as they failed the validation.

This PR resolves the issue by only requiring the from state if the country is US or Canada. 

**Steps to Reproduce**

1. Set store settings to an international address with no state (UK address for example)
2. Place a test order
3. Manually sync the order. The sync will fail and the logs will have the message indicating a required ship from field is missing.

**Expected Result**

After applying the PR, the order will sync correctly.

**Click-Test Versions**

- [X] Woo 4.4
- [X] Woo 4.3
- [X] Woo 4.2
- [X] Woo 4.1
- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2

**Specs Passing**

- [X] Woo 4.4
- [X] Woo 4.3
- [X] Woo 4.2
- [X] Woo 4.1
- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2

### DIFF
--- a/includes/class-taxjar-refund-record.php
+++ b/includes/class-taxjar-refund-record.php
@@ -57,8 +57,8 @@ class TaxJar_Refund_Record extends TaxJar_Record {
 			return false;
 		}
 
-		if ( empty( $data[ 'from_country' ] ) || empty( $data[ 'from_state' ] ) || empty( $data[ 'from_zip' ] ) || empty( $data[ 'from_city' ] ) ) {
-			$this->add_error( __( 'Refund failed validation - missing required ship from field on parent order.', 'wc-taxjar' ) );
+		if ( ! $this->has_valid_ship_from_address() ) {
+			$this->add_error( __( 'Order failed validation, missing required ship from field', 'wc-taxjar' ) );
 			return false;
 		}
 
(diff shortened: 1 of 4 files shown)


includes/class-taxjar-refund-record.php
```diff
@@ -57,8 +57,8 @@ public function should_sync() {
 			return false;
 		}
 
-		if ( empty( $data[ 'from_country' ] ) || empty( $data[ 'from_state' ] ) || empty( $data[ 'from_zip' ] ) || empty( $data[ 'from_city' ] ) ) {
-			$this->add_error( __( 'Refund failed validation - missing required ship from field on parent order.', 'wc-taxjar' ) );
+		if ( ! $this->has_valid_ship_from_address() ) {
+			$this->add_error( __( 'Order failed validation, missing required ship from field', 'wc-taxjar' ) );
```
fastdivision, line 61: @dallendalton Should this say `Refund failed validation` instead of `Order`?

dallendalton: Ahh yes you are correct. That was not an intentional change and I will get it updated.

PR #166, opened 2020-08-25
Fix incorrect trial on synchronized renewals

dallendalton:
When there was a subscription product set up to have synchronized renewals with no trial period, on the cart page a message would display on the item indicating it contained a 1 month free trial.

After testing it was determined that the responsible portion of code was no longer needed. This PR resolves the issue by removing that code.

It also adds an expected deprecated annotation to some of the unit tests for subscriptions. This ensures the tests will be able to pass on the most recent versions of WooCommerce. There is no way to do deprecated annotations based on the version of WooCommerce we are testing against so this will cause those tests to fail on older versions of Woo until we build a new way to create subscription orders in the test suite.

**Steps to Reproduce**

1. Create a subscription product with synchronized renewals
2. Add it to the cart and navigate to the cart page
3. On the line for that product, a message containing "with one month free trial" will be displayed even though no trial has been configured.

**Expected Result**

After applying the PR, the free trial portion of the message will no longer be added.

**Click-Test Versions**

- [X] Woo 4.4
- [X] Woo 4.3
- [X] Woo 4.2
- [X] Woo 4.1
- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2

**Specs Passing**

- [X] Woo 4.4
- [X] Woo 4.3
- [X] Woo 4.2
- [X] Woo 4.1
- [X] Woo 4.0
- [X] Woo 3.9
- [X] Woo 3.8
- [X] Woo 3.7
- [X] Woo 3.6
- [X] Woo 3.5
- [X] Woo 3.4
- [X] Woo 3.3
- [X] Woo 3.2

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -1133,9 +1133,6 @@ class WC_Taxjar_Integration extends WC_Settings_API {
 			// Get WC Subscription sign-up fees for calculations
 			if ( class_exists( 'WC_Subscriptions_Cart' ) ) {
 				if ( 'none' == WC_Subscriptions_Cart::get_calculation_type() ) {
-					if ( class_exists( 'WC_Subscriptions_Synchroniser' ) ) {
-						WC_Subscriptions_Synchroniser::maybe_set_free_trial();
-					}
 					$unit_price = WC_Subscriptions_Cart::set_subscription_prices_for_calculation( $unit_price, $product );
 				}
 			}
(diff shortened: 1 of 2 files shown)


dallendalton:
@fastdivision - Yep from all my testing that portion of code made no difference in any scenario (apart from creating the issue this PR resolves) for versions that we support. I looked back in the commit history and it was added when support for subscriptions was initially created. I believe it must have addressed something in either the older versions of Woo or of the subscriptions plugin.

PR #170, opened 2020-12-04
Subscription PTC Fix

dallendalton:
Previously there were periodic issues where PTCs were not being correctly applied to subscription and variable subscription products. This PR resolves the issues. 

It also includes a test to confirm full functionality in the future.

Note: This PR also includes a lot of changes to bring the changed files in line with our style guide. Comments will be added on the PR to indicate which changes are addressing the actual issue. I also recommend viewing the diff without showing whitespace differences.

**Steps to Reproduce**

1. Create a subscription variable product
2. At a PTC to that product
3. Add the product the cart and navigate to checkout. 
4. Check the tax calculation logs and see if the PTC is correctly sent to TaxJar.
5. You will see that the correct PTC was not sent to TaxJar

Note: Depending on the version of WooCommerce Subscriptions you have installed, this issue can be very hard to duplicate.

**Expected Result**

After applying the PR, the correct PTC will be applied.

**Click-Test Versions**

- [X] Woo 4.7
- [X] Woo 3.2

**Specs Passing**

- [X] Woo 4.7
- [X] Woo 3.2

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -1118,8 +1118,8 @@ class WC_Taxjar_Integration extends WC_Settings_API {
 		$line_items = array();
 
 		foreach ( $wc_cart_object->get_cart() as $cart_item_key => $cart_item ) {
-			$product = $cart_item['data'];
-			$id = $product->get_id();
+			$id = $cart_item[ 'data' ]->get_id();
+			$product = wc_get_product( $id );
 			$quantity = $cart_item['quantity'];
 			$unit_price = wc_format_decimal( $product->get_price() );
 			$line_subtotal = wc_format_decimal( $cart_item['line_subtotal'] );
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -9,1437 +9,1475 @@
 
 if ( ! class_exists( 'WC_Taxjar_Integration' ) ) :
 
-class WC_Taxjar_Integration extends WC_Settings_API {
+	class WC_Taxjar_Integration extends WC_Settings_API {
 
-	protected static $_instance = null;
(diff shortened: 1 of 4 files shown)


tests/specs/test-class-subscriptions.php
```diff
@@ -366,6 +366,47 @@ function test_correct_taxes_for_subscription_products_with_other_products_and_tr
 		}
 	}
 
+	/**
+	 * Tests that the correct PTC and tax are applied to subscription products
+	 * Tests problem in ISSUE-1782q
+	 *
+	 * @throws Exception
+	 */
+	function test_tax_for_exempt_subscription() {
```
dallendalton, line 418: This test confirms that the correct PTC is applied to a subscription product.

includes/class-wc-taxjar-integration.php
```diff
@@ -1118,8 +1118,8 @@ public function get_line_items( $wc_cart_object ) {
 		$line_items = array();
 
 		foreach ( $wc_cart_object->get_cart() as $cart_item_key => $cart_item ) {
-			$product = $cart_item['data'];
-			$id = $product->get_id();
+			$id = $cart_item[ 'data' ]->get_id();
```
dallendalton: These two lines contain the primary fix for the issue. Due to the way we cause the calculation methods to rerun, and the subscription plugins interaction with them, the tax class of the product was being altered before we performed the API requests. This fix ensures the correct data from the product is used.

fastdivision: Code change makes sense to me. Use `wc_get_product` instead of relying on `$cart_item['data']`.

fastdivision (approved):
Reviewed commented code and spec changes. Looks good to me 👍

codymullins (approved):

PR #171, opened 2020-12-04
Exempt Customers Not Syncing Fix

dallendalton:
There were multiple issues regarding syncing exempt customers to TaxJar.

1. If the customer was missing a piece of required data (normally a name), it would fail to sync. Later if the data was added, it still would fail to sync unless one of the exemption settings was also changed.

This was fixed by falling back to use the user's username if no name was present. Also, the validation logic was altered so that any change to the user could trigger a sync to TaxJar, and only if all the data that is sent to TaxJar was identical as the last sync would syncing be skipped.

2. Tax exemptions would not save or sync unless the user creating the exemption had the manage_woocommerce capability. By default, this capability is only assigned to admins and store manager roles. In some circumstances stores may have customer service roles that do not have that capability (it also controls changing woocommerce store settings).

To fix this the edit_user capability is now accepted as well.

Note: This PR also contains many fixes to style guide errors and warnings. Comments will be added to the PR to indicate which changes are related to the customer exemption fixes. I also recommend viewing the PR with white space differences turned off.

**Steps to Reproduce**

1. Create a customer without a first or last name.
2. Add an exemption type to the customer and save.
3. Check the customer sync logs, the customer will not have been synced.
4. Add a name to the customer.
5. Check the customer sync logs, the customer still will not have synced.

**Expected Result**

After applying the PR, the customer will have synced in step 3 (since it now falls back to username), and again in step 5.

**Click-Test Versions**

- [X] Woo 4.7
- [X] Woo 3.2

**Specs Passing**

- [X] Woo 4.7
- [X] Woo 3.2

### DIFF
--- a/includes/class-taxjar-customer-record.php
+++ b/includes/class-taxjar-customer-record.php
@@ -184,36 +184,8 @@ class TaxJar_Customer_Record extends TaxJar_Record {
 		$customer_data = array();
 
 		$customer_data[ 'customer_id' ] = $this->get_customer_id();
+		$customer_data[ 'name' ] = $this->get_customer_name();
 		$customer_data[ 'exemption_type' ] = $this->get_exemption_type();
-
-		$first_name = $this->object->get_shipping_first_name();
-		$last_name = $this->object->get_shipping_last_name();
-
-		if ( empty( $first_name ) ) {
-			$first_name = $this->object->get_billing_first_name();
-			if ( empty( $first_name ) ) {
-				$first_name = $this->object->get_first_name();
-			}
-		}
-
-		if ( empty( $last_name ) ) {
-			$last_name = $this->object->get_billing_last_name();
-			if ( empty( $last_name ) ) {
-				$last_name = $this->object->get_last_name();
-			}
-		}
-
-		if ( empty( $first_name ) && empty( $last_name ) ) {
-			$name = '';
-		} else if ( empty( $first_name ) ) {
-			$name = $last_name;
-		} else if ( empty( $last_name ) ) {
-			$name = $first_name;
-		} else {
-			$name = $first_name . ' ' . $last_name;
-		}
-		$customer_data[ 'name' ] = $name;
-
 		$customer_data[ 'exempt_regions' ] = $this->get_exempt_regions();
 
 		$country = $this->object->get_shipping_country();
@@ -259,6 +231,47 @@ class TaxJar_Customer_Record extends TaxJar_Record {
 		$customer_data = apply_filters( 'taxjar_customer_sync_data', $customer_data, $this->object );
 		$this->data = $customer_data;
 		return $customer_data;
+	}
+
+	/**
+	 * Retrieves the name of the customer.
+	 * Falls back to username if no shipping, billing or account names are available.
+	 *
+	 * @return string - Customer's name
+	 */
+	public function get_customer_name() {
+		$first_name = $this->object->get_shipping_first_name();
+		$last_name = $this->object->get_shipping_last_name();
+
+		if ( empty( $first_name ) ) {
+			$first_name = $this->object->get_billing_first_name();
+			if ( empty( $first_name ) ) {
+				$first_name = $this->object->get_first_name();
+			}
+		}
+
+		if ( empty( $last_name ) ) {
+			$last_name = $this->object->get_billing_last_name();
+			if ( empty( $last_name ) ) {
+				$last_name = $this->object->get_last_name();
+			}
+		}
+
+		if ( empty( $first_name ) && empty( $last_name ) ) {
+			$name = '';
+		} else if ( empty( $first_name ) ) {
+			$name = $last_name;
+		} else if ( empty( $last_name ) ) {
+			$name = $first_name;
+		} else {
+			$name = $first_name . ' ' . $last_name;
+		}
+
+		if ( empty( $name ) ) {
+			$name = $this->object->get_username();
+		}
+
+		return $name;
 	}
 
 	/**
--- a/includes/class-taxjar-customer-record.php
+++ b/includes/class-taxjar-customer-record.php
@@ -14,7 +14,7 @@ class TaxJar_Customer_Record extends TaxJar_Record {
 			$this->object = $object;
 		} else {
 			try {
-				$customer =  new WC_Customer( $this->get_record_id() );
+				$customer = new WC_Customer( $this->get_record_id() );
 				if ( $customer instanceof WC_Customer ) {
 					$this->object = $customer;
 				} else {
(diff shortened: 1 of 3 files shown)


includes/class-wc-taxjar-customer-sync.php
```diff
@@ -146,44 +146,111 @@ public function add_customer_meta_fields( $user ) {
...
 
+	/**
+     * Saves tax exemption user meta if necessary
+     *
+	 * @param $user_id - Id of user to update
+	 */
 	public function save_customer_meta_fields( $user_id ) {
 		if ( ! apply_filters( 'taxjar_current_user_can_edit_customer_meta_fields', current_user_can( 'manage_woocommerce' ), $user_id ) ) {
 			return;
 		}
```
dallendalton, line 165: Refactored this section to make the logic easier to follow. Also, it is no longer validating that the exemption settings on the user have changed at this point, instead it relies on the logic in the should_sync method of the record.

includes/class-wc-taxjar-customer-sync.php
```diff
@@ -146,44 +146,111 @@ public function add_customer_meta_fields( $user ) {
...
+			$exemption_settings_changed = true;
 		}
 
-		if ( $change ) {
+		if ( $exemption_settings_changed ) {
 		    do_action( 'taxjar_customer_exemption_settings_updated', $user_id );
         }
 	}
 
+	/**
```
dallendalton, line 172: The following four methods were added as part of the refactoring of the save_customer_meta_fields method. This methods were added to make the logic easier to follow.

includes/class-taxjar-customer-record.php
```diff
@@ -184,36 +184,8 @@ public function get_data_from_object() {
 		$customer_data = array();
 
 		$customer_data[ 'customer_id' ] = $this->get_customer_id();
+		$customer_data[ 'name' ] = $this->get_customer_name();
```
dallendalton: This portion was refactored into its own method to improve the readability of this method.

includes/class-taxjar-customer-record.php
```diff
@@ -261,6 +233,47 @@ public function get_data_from_object() {
 		return $customer_data;
 	}
 
+	/**
+	 * Retrieves the name of the customer.
+	 * Falls back to username if no shipping, billing or account names are available.
+	 *
+	 * @return string - Customer's name
+	 */
+	public function get_customer_name() {
```
dallendalton, line 254: This method contains the same logic as was pulled out of the get_data_from_object method. The only exception is that a fallback to the username was added.

includes/class-taxjar-customer-record.php
```diff
@@ -261,6 +233,47 @@ public function get_data_from_object() {
...
+			$name = '';
+		} else if ( empty( $first_name ) ) {
+			$name = $last_name;
+		} else if ( empty( $last_name ) ) {
+			$name = $first_name;
+		} else {
+			$name = $first_name . ' ' . $last_name;
+		}
+
+		if ( empty( $name ) ) {
```
dallendalton, line 282: If all the other potential name fields are empty, it now falls back to the username.

tests/specs/test-customer-sync.php
```diff
@@ -121,11 +123,11 @@ function test_customer_sync_validation() {
 		$should_sync = $record->should_sync();
 		$this->assertFalse( $should_sync );
 
-		// test no name
+		// test no name, falls back to username and should still sync
 		$record = new TaxJar_Customer_Record( $customer->get_id(), true );
 		$record->load_object();
 		$should_sync = $record->should_sync();
```
dallendalton, line 129: Test has been updated to now confirm that even if a first and last name is not present on the customer, the customer will still sync to TaxJar (due to the fallback to username).

includes/class-wc-taxjar-customer-sync.php
```diff
@@ -152,7 +152,8 @@ public function add_customer_meta_fields( $user ) {
 	 * @param $user_id - Id of user to update
 	 */
 	public function save_customer_meta_fields( $user_id ) {
-		if ( ! apply_filters( 'taxjar_current_user_can_edit_customer_meta_fields', current_user_can( 'manage_woocommerce' ), $user_id ) ) {
+		$has_permission = current_user_can( 'manage_woocommerce' ) || current_user_can( 'edit_users' );
```
dallendalton, line 161: Added the edit_users permission here.

includes/class-wc-taxjar-customer-sync.php
```diff
@@ -142,48 +148,110 @@ public function add_customer_meta_fields( $user ) {
...
+	/**
+	 * Checks if exempt regions have changed for a user
+	 *
+	 * @param $user_id - Id of user to check
+	 *
+	 * @return bool - Whether or not the exempt regions have changed
+	 */
+	public function have_exempt_regions_changed( $user_id ) {
+		$saved_exempt_regions = self::get_saved_exempt_regions( $user_id );
+		$new_exemption_type   = $this->get_posted_exempt_regions();
```
fastdivision, line 207: @dallendalton Naming tweak: `$new_exemption_type` -> `$new_exempt_regions`

fastdivision (approved):
Reviewed code and spec changes, looks good to me! Left some minor feedback on naming.

includes/class-wc-taxjar-customer-sync.php
dallendalton: Nice catch, this has been fixed.

includes/class-taxjar-customer-record.php
```diff
@@ -183,117 +195,142 @@ public function get_from_taxjar() {
...
+		$last_name  = $this->object->get_shipping_last_name();
+
+		if ( empty( $first_name ) ) {
+			$first_name = $this->object->get_billing_first_name();
+			if ( empty( $first_name ) ) {
+				$first_name = $this->object->get_first_name();
+			}
+		}
+
+		if ( empty( $last_name ) ) {
```
codymullins, line 265: Will `first_name` every be initially empty while `last_name` is populated? Don't know that we'd want to run in to a scenario where we use the first name of one person and the last name of another.

codymullins:
I'm not sure I understand, I see that `manage_woocommerce` is required but why `edit_user`?

dallendalton:
> I'm not sure I understand, I see that `manage_woocommerce` is required but why `edit_user`?

manage_woocommerce is a capability that gives a user power to edit site wide woocommerce settings. This capability is typically given to the admin and shop manager roles. So basically if a user with one of those adds an exemption for a customer it will cause the customer to be created in TaxJar.

This works fine for the majority of stores, as the users editing customers would generally have one of those two roles. However for larger merchants, the number of roles usually increases. For example a store might be setup with a customer service, and customer service manager roles. These roles would not be assigned the manage_woocommerce capability as they should not be allowed to change site wide settings. However they would be responsible for managing and editing customers. This is what the edit_users role controls.

So this change means a user with either of those two capabilities (edit_users or manage_woocommerce) could add exemptions to a customer and have them be synced to TaxJar.

includes/class-taxjar-customer-record.php
dallendalton: You're right, the logic here needed to be updated. I have changed it so it will only use the corresponding names (shipping, billing or account) together.

codymullins (approved):

PR #172, opened 2020-12-16
Handle non existent record queue table during transaction backfill

dallendalton:
There is a rare issue that occurs if a problem happens during plugin installation, that may cause the record queue database table to not be installed. In this case the user does not know there is an issue, however transaction sync will not ever work.

This PR fixes the issue by checking for this and adding an error message during backfill that indicates the problem and how to fix it.

**Steps to Reproduce**

1. Drop the record queue database table
2. Run transaction backfill (it needs to find at least one record to update).
3. A message will display indicating that records have been added to the queue, however the queue list view will be empty.

**Expected Result**

After applying the PR, the message should display that the database table was missing and give a suggestion for fixing it.

**Click-Test Versions**

- [X] Woo 4.8
- [X] Woo 3.6

**Specs Passing**

- [X] Woo 4.8
- [X] Woo 3.6


fastdivision (approved):
Reviewed code changes, LGTM 👍

codymullins (approved):

PR #175, opened 2021-01-22
Skip Tax Calc API Request For Zero Total Orders

dallendalton:
The plugin currently calculates tax even for orders with a $0 total. This is not necessary. This PR stops the plugin from sending an API request to TaxJar when the order total is $0. This applies to orders created through the cart, in the dashboard, subscription renewals and orders created through the Woo REST API.

**Steps to Reproduce**

1. Place a order with a $0 total
2. Look in the logs and you will see an API request was sent to TaxJar.

**Expected Result**

After applying the PR, no API request will have been sent for those orders.

**Click-Test Versions**

- [X] Woo 4.9
- [X] Woo 4.3

**Specs Passing**

- [X] Woo 4.9
- [X] Woo 4.3

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -986,6 +986,11 @@ if ( ! class_exists( 'WC_Taxjar_Integration' ) ) :
 		 */
 		public function calculate_backend_totals( $order_id ) {
 			$order      = wc_get_order( $order_id );
+
+			if ( ! $this->should_calculate_order_tax( $order ) ) {
+			    return;
+            }
+
 			$address    = $this->get_backend_address();
 			$line_items = $this->get_backend_line_items( $order );
 
@@ -1057,6 +1062,10 @@ if ( ! class_exists( 'WC_Taxjar_Integration' ) ) :
 				return $order;
 			}
 
+			if ( ! $this->should_calculate_order_tax( $order ) ) {
+			    return $order;
+            }
+
 			$this->calculate_order_tax( $order );
 
 			// must calculate tax on subscription in order for my account to properly display the correct tax
@@ -1075,14 +1084,13 @@ if ( ! class_exists( 'WC_Taxjar_Integration' ) ) :
 		 */
 		public function should_calculate_order_tax( $order ) {
 			$should_calculate = true;
-
-			// Cannot use $order->get_total() as it is not available yet when creating orders through the REST API
 			$total = 0;
-			foreach( $order->get_items( 'line_item', 'shipping', 'fee' ) as $item ) {
-				$total += $item->get_total();
+
+			foreach( $order->get_items( 'line_item', 'tax', 'shipping' ) as $item ) {
+				$total += floatval( $item->get_total() );
 			}
 
-			if ( $total === 0 ) {
+			if ( $total <= 0 ) {
 				$should_calculate = false;
 			}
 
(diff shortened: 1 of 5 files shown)


includes/class-wc-taxjar-integration.php
```diff
@@ -839,21 +839,42 @@ public function smartcalcs_cache_request( $json ) {
...
 			if ( doing_action( 'woocommerce_add_to_cart' ) ) {
-				return;
+				$should_calculate = false;
 			}
 
+			if ( floatval( $wc_cart_object->get_total( null ) ) === 0.0 ) {
+				$should_calculate = false;
+            }
+
+		    return apply_filters( 'taxjar_should_calculate_cart_tax', $should_calculate );
```
devinsays, line 864: What do you think about adding `$wc_cart_object` as an additional argument?

```
return apply_filters( 'taxjar_should_calculate_cart_tax', $should_calculate, $wc_cart_object );
```

That would allow someone to conditionally calculate tax based on the cart object.

includes/class-wc-taxjar-integration.php
```diff
@@ -1047,12 +1068,38 @@ public function calculate_renewal_order_totals( $order, $subscription, $type ) {
...
+			$total = 0;
+			foreach( $order->get_items( 'line_item', 'shipping', 'fee' ) as $item ) {
+				$total += $item->get_total();
+			}
+
+			if ( $total === 0 ) {
+				$should_calculate = false;
+			}
+
+			return apply_filters( 'taxjar_should_calculate_order_tax', $should_calculate );
```
devinsays, line 1089: What do you think about adding `$order` as an additional filter argument?

```
return apply_filters( 'taxjar_should_calculate_cart_tax', $should_calculate, $order );
```

That would allow someone to conditionally calculate tax based order.

devinsays (commented):
I like this approach, thanks for the quick movement on it.

I think the only downside is that the $subscription and $order still get saved in `calculate_renewal_order_totals`.

```
$order->calculate_totals();
$subscription->calculate_totals();
```
I believe this causes a database write even if no data is actually changing.

https://github.com/woocommerce/woocommerce/blob/9ebf70bf8dfcd13bdf1f3278ec5b841b2ef80da5/includes/abstracts/abstract-wc-order.php#L1668

What do you think about just this at the top of the `calculate_renewal_order_totals` method?

```
if ( ! $this->should_calculate_order_tax( $order ) ) {
    return $order;
}
```

includes/class-wc-taxjar-integration.php
dallendalton: Thanks for catching this, yep it should definitely have the cart object in that filter. We will make that change before it gets released.

includes/class-wc-taxjar-integration.php
dallendalton: Same here as with the cart filter, definitely a change we will make before release. Thanks!

dallendalton:
@devinsays - Thanks again for your feedback! I believe everything you mentioned should now be resolved.

devinsays:
@dallendalton Thanks, updates look great!

The only item not addressed was the question about `calculate_renewal_order_totals`. I was hoping to avoid the $order->calculate_totals(); and $subscription->calculate_totals(); calls if tax does not need to be calculated for the renewal order. Comment was here: https://github.com/taxjar/taxjar-woocommerce-plugin/pull/175#pullrequestreview-574585090

dallendalton:
@devinsays That issue was also resolved in the latest commits. Now we are calling the should_calculate_order_tax method prior to calculating tax on the renewal and the subscription and prior to recalculating the totals for those orders.  The should_calculate_order_tax method includes validation that the order total is not 0. Thanks!

fastdivision (approved):

cas27 (approved):

dbernheisel (approved):

PR #189, opened 2021-05-21
Refactor Order Calculation

dallendalton:
The current way the plugin performs tax calculation for orders (in the admin dashboard, through the API, and for subscription and renewal orders) had multiple known issues. It was also built in such a way that made testing the functionality very difficult. This PR refactors that functionality into individual modules that are all decoupled from each other. It also adds unit and integration tests to cover the new modules. All the issues fixed in this PR only affects the tax calculation performed on orders. It does not affect the cart tax calculation (the cart tax calculation does not have these issues). The following issues are fixed by this PR:

- Tax calculation was not being performed on orders with only a fee (no product line items) when created or edited in the admin dashboard or when created through the Woo REST API.
- Tax calculation was not being performed on orders with only a shipping item (no product or fee line items) when created or edited in the admin dashboard or when created through the Woo REST API.
- Incorrect tax rates may have been applied on orders when the TaxJar API returned different tax rates for products in the same order that used the same product tax class.
- In some order tax calculation scenarios, when the shipping tax rate differed from the line item tax rate (with no product tax class applied), the correct shipping tax rate was not being applied.
- The product tax class associated with fees was not being account for in order tax calculations.
- Subscription orders created through the Woo REST API may not have had the correct tax calculated.
- Customers that were set as VAT exempt still had tax calculated on orders created through the admin dashboard and through the Woo REST API.
- In some cases, subscription orders without a shipping address were not falling back to the billing address, and so did not have tax calculation performed.
- Customer exemptions weren't being applied on subscription orders created through the admin dashboard.
- When changing the customer on an order in the admin dashboard, the customer exemption from the newly selected customer was not applied in all scenarios.

In addition to these issues the following improvements were made:
- Improved logging in the order calculation process. This makes it easier to locate logs for a particular order and to determine the cause when tax is not calculated for the order.

**Click-Test Versions**

- [X] Woo 5.3
- [X] Woo 4.8

**Specs Passing**

- [X] Woo 5.3
- [X] Woo 4.8


includes/TaxCalculation/class-order-calculation-logger.php
```diff
@@ -0,0 +1,229 @@
...
+			return '';
+		}
+	}
+
+	/**
+	 * Format response message from TaxJar API.
+	 *
+	 * @param array $response Response from TaxJar API.
+	 *
+	 * @return string
```
poops, line 164: Will `format_response_message` and `format_response_body` return strings if the response is empty?

dallendalton: So when concatenating strings, PHP treats the value of a method with no return statement as an empty string. So these methods do work as intended. However I think we should be more consistent and use the same approach across all the related methods. I'll make that fix. Thanks!

includes/utilities/class-constants-manager.php
```diff
@@ -0,0 +1,106 @@
...
+	/**
+	 * Checks if constant is true.
+	 * First checks within manager and then globally.
+	 *
+	 * @param string $name The name of the constant.
+	 *
+	 * @return bool
+	 */
+	public static function is_true( $name ) {
+		return self::is_defined( $name ) && self::get_constant( $name );
```
poops, line 38: Should this also check that `self::get_constant( $name ) === true`?

cas27 (approved):

poops (approved):

PR #190, opened 2021-05-28
Support WooCommerce Cart and Checkout Blocks

dallendalton:
This PR enables our plugin to calculate tax when using the WooCommerce cart and checkout blocks.

**Steps to Reproduce**

1. Activate WooCommerce blocks plugin.
2. Add the cart/checkout block to a page.
3. Add a product to the cart and navigate to the page.
4. Tax will not be calculated.

**Expected Result**

After applying the PR, tax will be calculated in the cart and checkout blocks.

**Click-Test Versions**

- [X] Woo 5.3
- [X] Woo 4.8

**Specs Passing**

- [X] Woo 5.3
- [X] Woo 4.8

### DIFF
--- a/includes/class-taxjar-tax-calculation.php
+++ b/includes/class-taxjar-tax-calculation.php
@@ -8,6 +8,7 @@
  */
 
 use TaxJar\Tax_Calculator_Builder;
+use TaxJar\WooCommerce\TaxCalculation\Block_Flag;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
@@ -22,9 +23,9 @@ class TaxJar_Tax_Calculation {
 		// Cache rates for 1 hour.
 		$this->cache_time = HOUR_IN_SECONDS;
 
-
 		if ( TaxJar_Settings::is_tax_calculation_enabled() ) {
 			$this->init_hooks();
+			Block_Flag::init_hooks();
 			$this->update_tax_options();
 		}
 	}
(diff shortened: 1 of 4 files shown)


fastdivision (approved):
Looks good to me 👍 @dallendalton Are we able to test `should_calculate_on_page()` with filters, e.g. setting `woocommerce_is_checkout` filter or `DOING_AJAX` constant?

includes/class-taxjar-tax-calculation.php
```diff
@@ -204,7 +205,7 @@ public function should_calculate_cart_tax( $wc_cart_object ) {
 		$should_calculate = true;
 
 		// If outside of cart and checkout page or within mini-cart, skip calculations
```
poops, line 207: Should this comment be removed since `should_calculate_on_page`'s body is self-documenting?

poops (approved):
Minor suggestion on slightly outdated comment, but code LGTM!

includes/class-taxjar-tax-calculation.php
dallendalton: Good catch, I've removed it.

dallendalton:
> Looks good to me 👍 @dallendalton Are we able to test `should_calculate_on_page()` with filters, e.g. setting `woocommerce_is_checkout` filter or `DOING_AJAX` constant?

Yeah we can, I didn't think it was necessary though since the functions used in that method are all native WooCommerce ones and our logic is very simple.

PR #191, opened 2021-06-16
Minor performance optimization for transaction sync

dallendalton:
This is a minor performance improvement to the transaction sync process. Transaction sync works using action scheduler. Every twenty minutes a scheduled action is run that will process all records in the queue and create/update them in TaxJar. If more records remain than a singe job can process, the action will be scheduled to run again at the next possible time. 

In addition to the queue processor there is also a method running on the "init" hook that ensures there is always a scheduled action to process the queue. This method exists because there have been occurrences of the scheduled action getting cancelled or removed and causing no transactions to sync to TaxJar. This PR moves that method from the "init" hook to the "admin_init" hook. This should reduce the number of queries performed against the action scheduler table while still ensuring the the process queue job remains scheduled.

**Click-Test Versions**

- [X] Woo 5.4
- [X] Woo 4.9

**Specs Passing**

- [X] Woo 5.4
- [X] Woo 4.9

### DIFF
--- a/includes/class-wc-taxjar-transaction-sync.php
+++ b/includes/class-wc-taxjar-transaction-sync.php
@@ -33,7 +33,7 @@ class WC_Taxjar_Transaction_Sync {
 		$transaction_sync_enabled = isset( $this->taxjar_integration->settings['taxjar_download'] ) && 'yes' == $this->taxjar_integration->settings['taxjar_download'];
 
 		if ( $sales_tax_enabled || $transaction_sync_enabled ) {
-			add_action( 'init', array( __CLASS__, 'schedule_process_queue' ) );
+			add_action( 'admin_init', array( __CLASS__, 'schedule_process_queue' ) );
 			add_action( self::PROCESS_QUEUE_HOOK, array( $this, 'process_queue' ) );
 		}
 


poops (approved):

dbernheisel (approved):
https://developer.wordpress.org/reference/hooks/admin_init/
> Note, this does not just run on user-facing admin screens. It runs on admin-ajax.php and admin-post.php as well.
>
> This is roughly analogous to the more general ‘init’ hook, which fires earlier.
>
> admin_init is triggered before any other hook when a user accesses the admin area.
>
> This hook doesn’t provide any parameters, so it can only be used to callback a specified function.

https://developer.wordpress.org/reference/hooks/init/

> Most of WP is loaded at this stage, and the user is authenticated. WP continues to load on the ‘init’ hook that follows (e.g. widgets), and many plugins instantiate themselves on it for all sorts of reasons (e.g. they need a user, a taxonomy, etc.).
>
> If you wish to plug an action once WP is loaded, use the ‘wp_loaded’ hook


makes sense to me

dallendalton:
@dbernheisel - Thanks for adding that additional info, I probably should have included more with the initial PR description. Everything you mentioned is correct. The primary difference we are utilizing here is that "init" is run on every single request, for both the front end and the admin. Using "admin_init" will reduce the number of times a query is performed against the scheduled actions table. A high volume customer reported that with thousands of simultaneous users the action table was being locked up due to the number of queries.

Issue #193, opened 2021-07-21
Fatal error when store processes subscription renewal


alexrlangford:
I am encountering a fatal error in this plugin when a subscription renewal order is processed. The error in question appears to be the same as in issue #192, though the circumstances surrounding it are different.

One of the store's automatic subscription renewals failed, generating a notice in WordPress. When I tried to follow the documentation steps (https://docs.woocommerce.com/document/subscriptions/scheduled-action-errors/) and manually renew the subscription, I received the following PHP error:

> Fatal error: Uncaught Error: Call to a member function get_tax_rate() on bool in /nas/content/live/tkmgacademy/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php:151 Stack trace: #0 /nas/content/live/tkmgacademy/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php(111): TaxJar\Order_Tax_Applicator->get_product_line_item_tax_rate(3569, Object(WC_Order_Item_Product)) #1 /nas/content/live/tkmgacademy/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php(93): TaxJar\Order_Tax_Applicator->create_product_tax_rate(3569, Object(WC_Order_Item_Product)) #2 /nas/content/live/tkmgacademy/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php(82): TaxJar\Order_Tax_Applicator->create_rate_and_apply_to_product_line_item(3569, Object(WC_Order_Item_Product)) #3 /nas/content/live/tkmgacademy/wp-conten in /nas/content/live/tkmgacademy/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php on line 151

The only line item in this order is a variable subscription product, with no sales tax collected on it.

Some additional background is that this store was switched from AvaTax to TaxJar earlier this month. I tested on a staging site to see if this switch might be a cause of this issue and was not able to recreate it so it appears to be something else.

Plugin versions:

WordPress: 5.7.2
WooCommerce: 5.4.2
TaxJar: 3.2.10

Let me know if there's any other info I can provide to help resolve this issue.

dallendalton:
[...]

Issue #192, opened 2021-07-20
Recalculation Button causes Fatal Error


danieldcarimus:
We're working on a custom ecomm site.

PHP Fatal error:  Uncaught Error: Call to a member function get_tax_rate() on bool in /Users/danieldayton/Local Sites/pteus-com/app/public/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php:151
Stack trace:
#0 /Users/danieldayton/Local Sites/pteus-com/app/public/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php(111): TaxJar\Order_Tax_Applicator->get_product_line_item_tax_rate(458, Object(WC_Order_Item_Product))
#1 /Users/danieldayton/Local Sites/pteus-com/app/public/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php(93): TaxJar\Order_Tax_Applicator->create_product_tax_rate(458, Object(WC_Order_Item_Product))
#2 /Users/danieldayton/Local Sites/pteus-com/app/public/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php(82): TaxJar\Order_Tax_Applicator->create_rate_and_apply_to_product in /Users/danieldayton/Local Sites/pteus-com/app/public/wp-content/plugins/taxjar-simplified-taxes-for-woocommerce/includes/TaxCalculation/class-order-tax-applicator.php on line 151

We're getting this error whenever the recalculation button is clicked on the order admin screen.

We've dug into the plugin some. It seems the $line_items array is empty. We've tried deactivating all other plugins except payment gateway (Elavon), WooCommerce and Taxjar and we're still getting the same error.

Any recommendations on causes/debugging?

WC - 5.5.1
Taxjar - 3.2.10
WP - 5.7.2

dallendalton:
@danieldcarimus Thanks for submitting this issue. We are currently investigating it (along with https://github.com/taxjar/taxjar-woocommerce-plugin/issues/193) and will post an update here as soon as we have more information.

Is this error occurring when any subscription renewal is processed on your site or are there renewals that are going through correctly as well?

danieldcarimus:
[...]

PR #194, opened 2021-07-22
Fix error when no line items are present in tax details

dallendalton:
This PR resolves the issues found in https://github.com/taxjar/taxjar-woocommerce-plugin/issues/193 and https://github.com/taxjar/taxjar-woocommerce-plugin/issues/192

When a customer set up their TaxJar account and did not configure any nexus states, the API response for nexus regions would return an empty list. When the nexus list was empty there was logic in our plugin that would determine that we should calculate tax for all orders. 

When the api response was returned for these orders, it would indicate that no nexus was found and would not include the "breakdown" field in the response. This scenario was not handled by the plugin and would lead to a fatal PHP error being thrown when attempting to apply the tax response to an order.

**Steps to Reproduce**

1. Create/use a TaxJar account with no nexus states configured.
2. In the WooCommerce admin dashboard (with the above TaxJar account connected) manually create an order.
3. For the shipping and billing address use an address that is not in the state configured in the WooCommerce store settings. If the store setting state it used, it will be sent as the from state on the API request and the TaxJar API will appropriately mark it as having nexus. This will prevent the issue from being replicated.
4. Add a product to the order and click the "Recalculate Totals" button.
5. The AJAX request triggered by the button will not finish, and a fatal PHP error will be present in the logs.

**Expected Result**

After applying the PR, the recalculate totals call with finish as expect and no tax will be applied to the order since it does not have nexus.

**Click-Test Versions**

- [X] Woo 5.5
- [X] Woo 5.0

**Specs Passing**

- [X] Woo 5.5
- [X] Woo 5.0

### DIFF
--- a/includes/TaxCalculation/class-order-tax-applicator.php
+++ b/includes/TaxCalculation/class-order-tax-applicator.php
@@ -11,6 +11,7 @@ namespace TaxJar;
 
 use Automattic\WooCommerce\Utilities\NumberUtil;
 use WC_Tax, WC_Abstract_Order;
+use \Exception;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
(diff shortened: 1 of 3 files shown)


jakedjohnson:
lgtm!

jakedjohnson (approved):

fastdivision (approved):

danieldcarimus:
@dallendalton thanks! I thought that might be the case. We don't have access to the clients TaxJar config, but I'll let them know that's the issue. Appreciate the quick turnaroud.

PR #198, opened 2021-09-16
Int 1959 Cart Tax Calculation Logger

dallendalton:
This PR adds the cart tax calculation logger as part of the refactoring the cart tax calculation. It also refactors the order tax calculation logger to utilize the new Tax_Calculation_Results class in preparation for improving the log storage and UI.


poops (approved):

includes/TaxCalculation/class-tax-calculation-logger.php
```diff
@@ -0,0 +1,174 @@
+<?php
+/**
+ * TaxJar Tax_Calculation_Logger
+ *
+ * @package TaxJar
+ */
+
+namespace TaxJar;
+
+use Exception, WC_Logger_Interface;
```
sethobey, line 10: Looks like everywhere else each import has it's own "use" statement

sethobey (approved):
lgtm

PR #199, opened 2021-09-30
Cart Tax Applicator

dallendalton:
Continuing the work of refactor tax is calculated and applied to a cart, this PR includes the functionality for apply tax to the cart.

### DIFF
--- a/includes/TaxCalculation/class-cart-tax-applicator.php
+++ b/includes/TaxCalculation/class-cart-tax-applicator.php
@@ -1,0 +1,331 @@
+<?php
+/**
+ * Cart Tax Applicator
+ *
+ * Applies tax details to cart.
+ *
+ * @package TaxJar
+ */
+
+namespace TaxJar;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+use WC_Cart;
+use WC_Tax;
+use Exception;
+
+/**
+ * Class Cart_Tax_Applicator
+ */
+class Cart_Tax_Applicator extends Tax_Applicator {
+
+	/**
+	 * Cart to apply tax to.
+	 *
+	 * @var WC_Cart
+	 */
+	private $cart;
+
+	/**
+	 * Tax applied to items.
+	 *
+	 * @var array
+	 */
+	private $items_taxes = array();
+
+	/**
+	 * Tax applied to shipping.
+	 *
+	 * @var array
+	 */
+	private $shipping_taxes = array();
+
+	/**
+	 * Tax applied to fees.
+	 *
+	 * @var array
+	 */
+	private $fee_taxes = array();
+
+	/**
+	 * Cart_Tax_Applicator constructor.
+	 *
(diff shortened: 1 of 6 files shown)


includes/TaxCalculation/class-cart-tax-applicator.php
```diff
@@ -0,0 +1,331 @@
...
+
+	/**
+	 * Builds a WooCommerce structured tax rate.
+	 *
+	 * @param float  $rate Tax rate.
+	 * @param string $shipping_taxable yes if shipping is taxable, no if not.
+	 *
+	 * @return array[]
+	 */
+	private function build_woocoommerce_rate( $rate, $shipping_taxable = 'no' ): array {
```
poops, line 134: Typo in "woocoommerce"

includes/TaxCalculation/class-cart-tax-applicator.php
```diff
@@ -0,0 +1,331 @@
...
+	 * @param array  $item item in cart.
+	 * @param array  $wc_rate WooCommerce structure rate.
+	 *
+	 * @return array
+	 */
+	private function apply_line_subtotal_tax( $item_key, $item, $wc_rate ): array {
+		$item_subtotal  = wc_add_number_precision( $item['data']->get_price() * $item['quantity'], false );
+		$subtotal_taxes = WC_Tax::calc_tax( $item_subtotal, $wc_rate );
+		$subtotal_tax   = array_sum( array_map( array( $this, 'round_line_tax' ), $subtotal_taxes ) );
+		$this->cart->cart_contents[ $item_key ]['line_tax_data']     = array( 'subtotal' => wc_remove_number_precision_deep( $subtotal_taxes ) );
```
poops, line 158: Do you think there's any benefit in having the `line_tax_data` array instantiated if needed in this method & `apply_line_total_tax`, so the order of the function calls don't matter?

poops (approved):
Minor comments, but LGTM

includes/TaxCalculation/class-cart-tax-applicator.php
sethobey: I see how it is procedurally safe as-is, but I think @poops has a point.

If it was something like:

`$this->cart->cart_contents[ $item_key ]['line_tax_data']['subtotal'] = wc_remove_number_precision_deep( $subtotal_taxes );
 `

as you do on line 175 below, then if these methods were called in a different order, it wouldn't be destructive to the 'total' value set in `apply_line_total_tax`

sethobey (approved):

includes/TaxCalculation/class-cart-tax-applicator.php
dallendalton: Good catch, fixed!

includes/TaxCalculation/class-cart-tax-applicator.php
dallendalton: I agree this is a better way and have fixed it. Thanks!

PR #207, opened 2021-11-24
Save Calculation Results to Cart and Order

dallendalton:
In order to allow merchants to have more visibility into if and when TaxJar calculated tax on a cart/order and why we may not have calculated tax (for instance no nexus, etc.) we want to surface the tax calculation results to the order page in WooCommerce admin. This PR represents the first step to accomplish that.

This PR stores a JSON representation of the calculation results on the cart during the checkout process (and on recurring carts as well). When an order is created from a cart or a subscription is created from a recurring cart during checkout the JSON representation is then saved to the order/subscription as meta data.

**Click-Test Versions**

- [X] Woo 5.9

**Specs Passing**

- [X] Woo 5.9

### DIFF
--- a/includes/class-taxjar-tax-calculation.php
+++ b/includes/class-taxjar-tax-calculation.php
@@ -39,7 +39,7 @@ class TaxJar_Tax_Calculation {
 
 		add_action( 'woocommerce_order_after_calculate_totals', array( $this, 'maybe_calculate_order_taxes' ), 10, 2 );
 		add_action( 'woocommerce_checkout_create_order', array( $this, 'persist_cart_calculation_results_to_order'), 10, 1 );
-		add_action( 'woocommerce_checkout_subscription_created', array( $this, 'persist_recurring_cart_calculation_results_to_subscription'), 10, 3 );
+		add_action( 'woocommerce_checkout_create_subscription', array( $this, 'persist_recurring_cart_calculation_results_to_subscription'), 10, 4 );
 	}
 
 	public function persist_cart_calculation_results_to_order( $order ) {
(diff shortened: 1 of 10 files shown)


includes/class-taxjar-tax-calculation.php
```diff
@@ -38,6 +38,18 @@ public function init_hooks() {
...
+		add_action( 'woocommerce_checkout_create_order', array( $this, 'persist_cart_calculation_results_to_order'), 10, 1 );
+		add_action( 'woocommerce_checkout_create_subscription', array( $this, 'persist_recurring_cart_calculation_results_to_subscription'), 10, 4 );
+	}
+
+	public function persist_cart_calculation_results_to_order( $order ) {
+		$calculation_results = WC()->cart->tax_calculation_results;
+		$order->update_meta_data( '_taxjar_tax_result', $calculation_results );
+	}
+
+	public function persist_recurring_cart_calculation_results_to_subscription(  $subscription, $posted_data, $order, $recurring_cart ) {
```
sethobey, line 50: At first it looked like unused inputs here, but I'm guessing these "extra" params are necessary inputs when responding to the `woocommerce_checkout_create_subscription` hook?

sethobey (approved):
One Q, but LGTM

includes/class-taxjar-tax-calculation.php
dallendalton: Yep, in order to access the recurring cart parameter we have to include all the ones that come before it. These are declared when the hook is fired so not anything we can change.

jordan-k-johnson (approved):

PR #208, opened 2021-12-02
Admin Order Meta Box

dallendalton:
This PR introduces a meta box on orders in the admin dashboard. This meta box surfaces data from the TaxJar calculation and sync processes. This includes:

- The calculation status of the order. Whether or not TaxJar calculated tax on the order in realtime, and if not the reason why.
- The sync status of the order. If the order has been synced to TaxJar and the last sync time.
- Additional meta data. This includes the JSON request and response from the calculation process.

image

image

image

**Click-Test Versions**

- [X] Woo 5.9
- [X] Woo 5.4

**Specs Passing**

- [X] Woo 5.9
- [X] Woo 5.4

### DIFF
--- a/includes/js/wc-taxjar-order.js
+++ b/includes/js/wc-taxjar-order.js
@@ -61,7 +61,13 @@ jQuery( document ).ready( function() {
 		},
 
 		copy_success: function() {
-			alert('Copied to clipboard.');
+			jQuery( this ).tipTip({
+				'attribute':  'data-tip',
+				'activation': 'focus',
+				'fadeIn':     50,
+				'fadeOut':    50,
+				'delay':      0
+			}).focus().focus();
 		}
 	};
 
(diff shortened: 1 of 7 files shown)


includes/js/wc-taxjar-order.js
```diff
@@ -34,4 +34,36 @@ jQuery( document ).ready( function() {
...
+
+		copy_response_json: function( e ) {
+			wcClearClipboard();
+			wcSetClipboard( jQuery('#advanced_order_data .response-json .accordion-section-content pre').text(), jQuery( this ) );
+			e.preventDefault();
+			e.stopPropagation();
+		},
+
+		copy_success: function() {
+			alert('Copied to clipboard.');
```
sethobey, line 64: Definitely agree some visual indicator is necessary, but having to click a second button in the alert popup was kind of cumbersome. Although alternative solution probably not worth the development time.

sethobey (approved):

jordan-k-johnson (approved):

includes/js/wc-taxjar-order.js
dallendalton: The alert notification has now been changed to a tool tip. Thanks!

PR #210, opened 2021-12-09
Filter Invalid PTCs

dallendalton:
The TaxJar plugin parses PTCs (Product Tax Codes) from a given products tax class. These tax classes are natively functionality to WooCommerce which means that many invalid codes are sent to TaxJar when creating transactions. While this doesn't prevent the transaction from being created, it can create confusion for customers and potential confusion for filings.

This PR adds additional filtering to PTC parsing and ensures that only PTCs that are in a valid format are sent to TaxJar.

**Click-Test Versions**

- [X] Woo 5.9
- [X] Woo 5.4

**Specs Passing**

- [X] Woo 5.9
- [X] Woo 5.4

### DIFF
--- a/includes/class-taxjar-tax-calculation.php
+++ b/includes/class-taxjar-tax-calculation.php
@@ -198,7 +198,12 @@ class TaxJar_Tax_Calculation {
 			$tax_code = end( $tax_class );
 		}
 
-		return strtoupper( $tax_code );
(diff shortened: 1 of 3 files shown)


sethobey (approved):
lgtm

jordan-k-johnson (approved):

PR #219, opened 2022-02-08
Refactor cart shipping method tax application

sethobey:
This PR adds refactors shipping tax application to resolve double-taxation issue on packages cause by the same shipping method being present in session more than once.

### Steps to Reproduce

- Add simple taxable product to cart
- Navigate to checkout page to trigger tax calculation
- Observe calculated tax value is inflated

### Expected Result

After applying the PR, package-index will be used to retrieve shipping method from session, avoiding duplicate tax application.


dallendalton (approved):

jordan-k-johnson (approved):
LGTM

PR #222, opened 2022-03-10
Improves stored tax rate precision

### LINKED ISSUE #221: Possible regression, "Save Tax Rates" setting saves invalid rates for Postcode/ZIP/City
I am not 100% if this is a bug but from what I can tell during a refactor in #203 it seems that this line https://github.com/taxjar/taxjar-woocommerce-plugin/blob/9cd2c1155e24dd97753c44f1634a4743ee31ba63/includes/TaxCalculation/class-cart-tax-applicator.php#L105 was added that saves the calculated rate, not the rate from the upstream API, to woocommerce. This causes the tax rates in Woocommerce to be set for the ZIP/Postcode for whatever the calculated rate is for that line item at the time of calculation. It also saves a invalid `rate_percent` order item meta value to the order items.

image

In the screenshot above you can see the rates being saved are not correct for the areas that it is listed for. What I noticed is that it correctly gets the rate from the upstream api and then calculates the taxes for product/line item correctly. But after that it tries to recalculate the rate from the taxes applied and the total for the item which causes a loss of precision to occur and a less precise/invalid value to be stores as the tax rate. 

https://github.com/taxjar/taxjar-woocommerce-plugin/blob/9cd2c1155e24dd97753c44f1634a4743ee31ba63/includes/TaxCalculation/class-cart-tax-applicator.php#L81

 If I use https://www.taxjar.com/sales-tax-calculator the tax jar sales tax calculator you can see at the time of writing this the rate for Orangevale, CA is 7.75%
image

Where this trips up my customer is that this tax rate is added by Woocommerce to the order/order items as the `rate_percent` prop/order item meta key.

image

 These values were being sent to a third party system and up until #203 were the right/precise values based on the API responses from TaxJar. Now this upstream system does not accept these values as valid as it uses a tax lookup service and the rates are not correct for the areas/zip codes of the order.

Please let me know if this doesn't make sense/if you need more information on this.

Thanks,
Justin Welenofsky

sethobey:
Closes #221  

During tax calculation, the TaxJar plugin correctly stores the location's rate from the TaxJar API, but then overwrites the stored rate when calculating cart and order line subtotals with a "reconstructed" `$applied_rate` that is reverse-engineered from the line total. Rounding line item tax amounts in previous steps can lead to a loss of float precision when reconstructing the tax rate.

This PR adds an additional check for the existence of line item tax directly on the tax detail object before attempting to reconstruct an applied rate from item totals.

**Click-Test Versions**

- [X] Woo 5.9

**Specs Passing**

- [ ] Woo 5.9

### DIFF
--- a/includes/TaxCalculation/class-cart-tax-applicator.php
+++ b/includes/TaxCalculation/class-cart-tax-applicator.php
@@ -78,7 +78,17 @@ class Cart_Tax_Applicator extends Tax_Applicator {
 
 		foreach ( $this->cart->get_cart() as $item_key => $item ) {
 			$total_taxes           = $this->apply_line_total_tax( $item_key, $item );
-			$applied_rate          = empty( $item['line_total'] ) ? 0.0 : reset( $total_taxes ) / wc_add_number_precision( $item['line_total'] );
+			$tax_detail_key		   = $item['product_id'] . '-' . $item_key;
+			$tax_detail_line_item  = $this->tax_details->get_line_item($tax_detail_key);
+
+			if ( empty( $item['line_total'] ) ) {
+				$applied_rate = 0.0;
+			} elseif ( $tax_detail_line_item ) {
+				$applied_rate = $tax_detail_line_item->get_tax_rate();
+			} else {
+				$applied_rate = reset( $total_taxes ) / wc_add_number_precision( $item['line_total'] );
+			}
+
 			$subtotal_taxes        = $this->apply_line_subtotal_tax( $item_key, $item, $applied_rate );
 			$merged_subtotal_taxes = $this->merge_tax_arrays( $merged_subtotal_taxes, $subtotal_taxes );
 			$merged_total_taxes    = $this->merge_tax_arrays( $merged_total_taxes, $total_taxes );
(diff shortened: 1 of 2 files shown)


dallendalton (changes requested):
This approach is not going to work in all scenarios. There are jurisdictions that have threshold exemptions where directly using the combined_tax_rate returned by the TaxJar API would cause an incorrect tax subtotal to be applied to the item.

welenofsky:
One thing I am wondering is if we can do away with the saving of tax rates in the function `build_line_tax_from_rate` all together. I can see 3 usages across the code base and 2 of those 3 seem to use a calculated rate based on the taxes applied to the line item. It seems that 2 of those calls we may not want the tax rate created in woo due, the 3rd I am not certain. 

As a temporary patch I have added an optional param `$prevent_save` to the `build_line_tax_from_rate` method and passed that in through the instances where the tax rate is being reconstructed. 

https://github.com/three29/taxjar-woocommerce-plugin/commit/23924cb10d6057716b7705751f844e263bd2cb8b

If this is an acceptable solution I could PR this.

dallen-stripe:
@welenofsky I don't think that solution actually resolves the root issue. That change would prevent the plugin from creating any rates in the WooCommerce tax table except for the rates applied to fee items. While creating rates in the tax table does not actually affect the amount of tax applied to a cart/order, some customers do use it for reporting purposes. For those that don't want to create rates at all they can disable the save rates setting.

We are currently investigating changes that would not affect the amount applied to a line item but would resolve the root issue of a rate row being created with an incorrect rate %.

Thanks again for submitting this issue and working with us to get it solved!

PR #224, opened 2022-03-17
Prevent calculated tax rates from being saved to database

### LINKED ISSUE #221: Possible regression, "Save Tax Rates" setting saves invalid rates for Postcode/ZIP/City

welenofsky:
Closes #221 

This PR adds an optional parameter to Tax_Builder::build_line_tax_from_rate to prevent saving some tax rates to WooCommerce. This will prevent saving any tax rates that are calculated based on taxes applied to line items which are used as variables of other calculations.  Another solution to this would be to expose the private method `build_woocommerce_rate` but I wanted to get your feedback on this. I have ran this patch both in my staging environment and production now for a few days and everything has been going smooth.

## Here are some photos of the production data after the application of the patch.

taxjar-api-rate
**This is the tax rate for this orders ZIP code from the [taxjar sales tax calculator](https://www.taxjar.com/sales-tax-calculator).** 

rates-saved-db-rate_percent-taxjar
**This is the tax rate saved to the order item which is the rate percent my customer uses to send to an upstream provider as explained in #221. It now matches the correct value from taxjar sales tax calculation.**

rates-saved-db-woocommerce
**Here are various rates that have been calculated over the course of a few days for taxable orders. All of which are now matching the taxjar sales tax calculator values.**

**Click-Test Versions**

- [x] Woo 5.9

**Specs Passing**

- [ ] Woo 5.9

### DIFF
--- a/includes/TaxCalculation/class-cart-tax-applicator.php
+++ b/includes/TaxCalculation/class-cart-tax-applicator.php
@@ -102,7 +102,7 @@ class Cart_Tax_Applicator extends Tax_Applicator {
 	private function apply_line_subtotal_tax( $item_key, $item, $applied_rate ): array {
 		$item_subtotal  = wc_add_number_precision( $item['line_subtotal'], false );
 		$tax_class      = $item['data']->get_tax_class();
-		$subtotal_taxes = $this->tax_builder->build_line_tax_from_rate( $applied_rate, $item_subtotal, $tax_class );
+		$subtotal_taxes = $this->tax_builder->build_line_tax_from_rate( $applied_rate, $item_subtotal, $tax_class, true );
 		$subtotal_tax   = array_sum( array_map( array( $this, 'round_line_tax' ), $subtotal_taxes ) );
 		$this->cart->cart_contents[ $item_key ]['line_tax_data']['subtotal'] = wc_remove_number_precision_deep( $subtotal_taxes );
 		$this->cart->cart_contents[ $item_key ]['line_subtotal_tax']         = wc_remove_number_precision( $subtotal_tax );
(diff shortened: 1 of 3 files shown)


dallendalton:
Closing in favor of https://github.com/taxjar/taxjar-woocommerce-plugin/pull/225

PR #225, opened 2022-03-23
Separate persisting woo rates from tax application

dallendalton:
This PR resolves https://github.com/taxjar/taxjar-woocommerce-plugin/issues/221. 

The TaxJar API returns an amount to collect for each line item. This amount to collect is used to apply tax to line item total (after discounts have been applied). Tax must also be applied to the line item subtotal (prior to discounts). In order to get the amount to apply to the subtotal, an "applied rate" is determined by dividing the amount collectable by the line item total. This rate was then applied to the subtotal.

When the setting was enabled to save tax rates into the WooCommerce tax rate table, a rate was then created using the "applied rate". This rate overwrote the rate created while applying tax to the line total, and would be slightly different than the actual rate returned by the TaxJar API.

In order to resolve this issue, this PR separates the persisting the rate in the WooCommerce tax table from the actual application of tax to the line items. Now only a single rate will be created for each line item and the correct rate percentage will be persisted. This refactor also makes it easier to follow the logic of what is occurring during the tax application process.

Credit to @welenofsky for discovering this issue and pointing us in the right direction for a fix!

**Steps to Reproduce**

1. Enable the "save rates" and the "debug log" settings.
2. Add an item with an odd amount (for example $117) to the cart and proceed through the checkout process. The shipping address for the cart must be set to an area where nexus has been configured.
3. View the rate created in the WooCommerce tax table and the rate returned by the TaxJar API (can be seen in the logs) for the line item. They will be slightly different. 

**Expected Result**

After applying the PR, in step 3 above the two rates will be identical.

**Click-Test Versions**

- [X] Woo 6.3
- [X] Woo 5.8

**Specs Passing**

- [X] Woo 6.3
- [X] Woo 5.8

### DIFF
--- a/includes/TaxCalculation/class-cart-tax-applicator.php
+++ b/includes/TaxCalculation/class-cart-tax-applicator.php
@@ -77,9 +77,18 @@ class Cart_Tax_Applicator extends Tax_Applicator {
 		$merged_total_taxes    = array();
 
 		foreach ( $this->cart->get_cart() as $item_key => $item ) {
-			$total_taxes           = $this->apply_line_total_tax( $item_key, $item );
+			$tax_detail_line_key = $item['data']->get_id() . '-' . $item_key;
+			$tax_details_line_item = $this->tax_details->get_line_item( $tax_detail_line_key );
+
+			$rate_id = $this->tax_builder->build_woocommerce_tax_rate(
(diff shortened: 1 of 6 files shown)


sethobey (approved):
Tested locally and LGTM

jordan-k-johnson (approved):

Issue #228, opened 2022-05-13
Ignores discount from WooCommerce PDF Product Vouchers


rlunden:
When TaxJar calculates its taxes it ignores the discount when applying a coupon created with the WooCommerce PDF Product Vouchers plugin so the total is unchanged as if the coupon never was applied.
![image](https://user-images.githubusercontent.com/7823044/168237666-fbcc1839-d987-4617-87c5-5dbbb9c0c405.png)

dallendalton:
Thanks for submitting this issue. We are currently investigating it and will post an update once we have one.

rlunden:
Thanks @dallendalton. Any progress?

dallendalton:
@rlunden Thanks for your patience on this. We now have a branch up for testing a fix for this issue: 

https://github.com/taxjar/taxjar-woocommerce-plugin/tree/product-pdf-voucher-support

Could you confirm this resolves the issue?

Thanks!

rlunden:
@dallendalton It resolves the scenario when you have tax, but instead now it doesn't work when there is no tax:
![image](https://user-images.githubusercontent.com/7823044/173780751-d213a3b3-35f4-4927-b13f-40e97aba5b7f.png)

dallendalton:
Thanks for testing this! We are working on an update to the fix to resolve this and will hopefully have it up for testing tomorrow.

dallen-stripe:
We have updated that branch and it fixes this if you would like to test again. Thanks!

dallendalton:
We have updated that branch and it fixes this if you would like to test again. Thanks!

dallendalton:
This issue has been resolved in the latest release. Thanks again!

PR #232, opened 2022-06-14
Fix conflict with WooCommerce PDF Product Vouchers plugin

dallendalton:
This PR resolves the incorrect totals in cart while using the TaxJar and WooCommerce PDF Product Vouchers plugins as reported in https://github.com/taxjar/taxjar-woocommerce-plugin/issues/228.

**Steps to Reproduce**

1. Add an item to the cart and apply a multi use product voucher to the cart.
2. The totals will be incorrect as the voucher won't be factored into the total.

**Expected Result**

After applying the PR, the totals will be as expected.

**Click-Test Versions**

- [X] Woo 6.6
- [X] Woo 6.0

**Specs Passing**

- [X] Woo 6.6
- [X] Woo 6.0

### DIFF
--- a/includes/compatibility/class-module-loader.php
+++ b/includes/compatibility/class-module-loader.php
@@ -20,6 +20,7 @@ class Module_Loader {
 	private $extensions = array(
 		'\TaxJar\WooCommerce_Gift_Cards',
 		'\TaxJar\WooCommerce_Smart_Coupons',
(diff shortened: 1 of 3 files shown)


sethobey (approved):
lgtm

jordan-k-johnson (approved):

PR #255, opened 2024-07-04
Fix for two HPOS related issues and a small improvement for the readme

ahumulescu:
**This PR:**
- Fixes a bug where the tax would still be calculated for an exempt customer if the order was created by and admin and HPOS was active
- Fixes a secondary bug, where the TaxJar meta boxes would not be displayed in each order's individual page if HPOS was active
- Fixes a small inconsistency in the readme.txt file, which was introduced in version 2.1.0. The [Changelog](https://wordpress.org/plugins/taxjar-simplified-taxes-for-woocommerce/#developers) section does not have the release version section headers bold since then.

**Steps to Reproduce**
1. Go to WordPress admin -> WooCommerce -> Settings -> Advanced -> Features -> enable HPOS
2. Go to Users -> Select a user -> and set the Exemption Type to 'Government'
3. Go to Orders -> Add order -> Select the user mentioned above, add a product and click 'Recalculate'

**Expected Result**
- The tax not being calculated
- The meta box being displayed

**After applying the PR, this should happen.**
- The tax will not be calculated for the exempt customers
- The meta box will be displayed in each order's individual page

**Click-Test Versions**
- [ ] Woo 9.0

**Specs Passing**
- [ ] Woo 9.0

### DIFF
--- a/includes/admin/class-admin-meta-boxes.php
+++ b/includes/admin/class-admin-meta-boxes.php
@@ -9,6 +9,8 @@
 
 namespace TaxJar;
 
+use Automattic\WooCommerce\Utilities\OrderUtil;
+
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
 }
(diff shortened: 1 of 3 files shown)


alis-stripe (approved):
Thanks for the PR. Click tested and everything works as expected.

PR #266, opened 2024-11-22
Update readme.txt to update tested version of WP to 6.7

isehrob-stripe:
WordPress 6.7 has been releases on November 12, 2024. But the plugin was not tested for this new version.

image

I've performed tests with this latest version 6.7 of WP and confirmed that the plugin works as expected. Thus updating the readme.txt to reflect this fact. Based on the email linked in the ticket there's nothing else required from our side.

For more details checkout RUN_TJPLAT-1946.


saville-stripe (approved):
LGTM

PR #267, opened 2025-03-10
PHP 8.x Compliance

meyerauslander:
Define some class properties that are currently dynamic because [PHP has deprecated dynamic class props](https://www.php.net/manual/en/migration82.deprecated.php#migration82.deprecated.core.dynamic-properties).
```


meyerauslander:
Commits not associated to user

PR #270, opened 2025-08-05
Fix PHP 8.2+ dynamic property deprecation warnings

corys-stripe:
- Add explicit property declarations to prevent dynamic property warnings:
  - Add $plugin_dir property to TaxJar_WC_Unit_Tests_Bootstrap class
  - Add $log property to WC_Taxjar_Integration class
  - Add $tj property to TJ_WC_Actions test class

- Fix WooCommerce Subscriptions file loading with existence check in bootstrap.php to prevent file not found warnings

- Replace problematic shipping total assignments with proper WooCommerce methods:
  - Replace WC()->shipping->shipping_total = X with WC()->cart->set_shipping_total(X)
  - Remove redundant shipping_total assignment in woocommerce-helper.php

### DIFF
--- a/includes/class-wc-taxjar-integration.php
+++ b/includes/class-wc-taxjar-integration.php
@@ -38,6 +38,11 @@ if ( ! class_exists( 'WC_Taxjar_Integration' ) ) :
 		public $debug;
 
 		/**
+		 * @var WC_Logger
+		 */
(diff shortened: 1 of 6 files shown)


culley-stripe (approved):
✅

PR #298, opened 2026-03-23
Fix: Skip negative fee line items to prevent gift card 400 errors

kkolk-stripe:
When plugins like WooCommerce Gift Cards apply discounts as negative fee line items, the TaxJar API rejects the request with a 400 error: "amount must be equal to the sum of line items and shipping, excluding sales tax." This happens because the API converts all line item amounts to absolute values, inflating the total beyond the order amount.

This fix skips fee line items with negative amounts across all three code paths that build API requests: transaction sync, cart tax calculation, and order tax calculation.

The existing gift card compatibility module (`class-woocommerce-gift-cards.php`) correctly adjusts the top-level `amount` field but did not filter the negative fee from `line_items`, causing a double-counting mismatch.

**Steps to Reproduce**

1. Install WooCommerce Gift Cards plugin
2. Create a gift card that covers the full order total ($0 due)
3. Attempt checkout — fails with "no payment method provided"
4. Check API logs — 400 error on v2/transactions

**Expected Result**

After applying the PR, negative fee line items (gift cards, store credits) are excluded from API requests. The `amount` field remains correctly adjusted by the compatibility module, and the equation `amount == sum(line_items) + shipping` holds.

**Jira:** RUN_TJPLAT-3033
**HelpScout:** Conversation 3209826668

**Specs Passing**

- Added new test cases to cover this to existing test suite

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -80,10 +80,12 @@ steps:
     artifact_paths:
       - "phpcs-report.xml"
 
-  # Stage 3: PHPUnit Tests - Parallel execution with unique transaction IDs
+  # Stage 3: PHPUnit Tests - Serial execution across WC versions
   - label: ":test_tube: PHPUnit - WC {{matrix}}"
     key: "phpunit"
     depends_on: "lint"
+    concurrency: 1
+    concurrency_group: "taxjar-woocommerce-phpunit-$BUILDKITE_BUILD_ID"
     matrix:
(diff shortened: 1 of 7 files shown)


kkolk-stripe:
@asmolentzov @corys-stripe Looking to close this or merge this if either of you have a moment to look at it so I can close out https://jira.corp.stripe.com/browse/RUN_TJPLAT_INFRA-1

